PR #1, opened 2014-12-31
Slight refactor in ast.js

eganj:


EisenbergEffect:
Unfortunately, I can't take this PR. The refactoring does simplify the code, however it changes its behavior in that it forces and evaluation of the right side of the logical expression even if the left side is part of an AND and evaluates to false.

PR #12, opened 2015-01-31
fix(binding): make global functions accessible in templates

### LINKED ISSUE #11: Global Functions not available to be called in template
The following will fail.

```html
<h2>${encodeURI(heading)}</h2>
```

Is this by design or a bug?

Stack trace is:
```
Potentially unhandled rejection [1] Error: encodeURI is not a function
    at ensureFunctionFromMap (http://localhost:9000/jspm_packages/github/aurelia/binding@0.3.2/system/ast.js:58:13)
    at CallScope.evaluate (http://localhost:9000/jspm_packages/github/aurelia/binding@0.3.2/system/ast.js:652:22)
    at CallScope.connect (http://localhost:9000/jspm_packages/github/aurelia/binding@0.3.2/system/ast.js:692:29)
    at InterpolationBinding.connect (http://localhost:9000/jspm_packages/github/aurelia/templating-binding@0.8.3/system/binding-language.js:232:35)
    at InterpolationBinding.bind (http://localhost:9000/jspm_packages/github/aurelia/templating-binding@0.8.3/system/binding-language.js:201:22)
    at View.bind (http://localhost:9000/jspm_packages/github/aurelia/templating@0.8.6/system/view.js:69:29)
    at RouterView.swap (http://localhost:9000/jspm_packages/github/aurelia/templating-router@0.9.2/system/router-view.js:85:49)
    at http://localhost:9000/jspm_packages/github/aurelia/router@0.5.1/system/navigation-context.js:72:37
    at Array.forEach (native)
    at http://localhost:9000/jspm_packages/github/aurelia/router@0.5.1/system/navigation-context.js:71:28es6-module-loader.src.js:139 (anonymous function)es6-module-loader.src.js:166 fes6-module-loader.src.js:188 i
```

AshleyGrant:
fixes #11 

I'm not sure if this is a fix for a bug or changing the expected/desired behavior, but it's here for your review.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -970,7 +970,7 @@ function autoConvertAdd(a, b) {
 }
 
 function ensureFunctionFromMap(obj, name){
-  var func = obj[name];
+  var func = obj[name] || window[name];
 
   if (typeof func === 'function') {
     return func;


EisenbergEffect:
I have to think about this. I'm not sure if it's a good idea. I'm leaning against it. What is the use case for this? Why can't the code be placed in a view-model?

AshleyGrant:
It came from a question asked in the gitter. I just looked at the stack
trace on saw how to fix it. I wasn't sure if it was a good idea, so I just
created the PR for you to decide :-)

As far as the use case, you'll have to discuss with the use in gitter.

tomasbonco:
Hello,

Example of use case:
```
<div repeat.for="url of urls">
    <a href="#/url/${ encodeURIComponent( url ) }"> ${ url } </a> 
</div>
```

I think it would also add support for underscore/lodash functions then (without hacking). It would bring similar behaviour as Angluar's filters. Example using lodash:
```
${ _.camelCase( title ) }
```

So main focus is to add filters.

EisenbergEffect:
Sure. Aurelia does have ValueConverters though...which do the same thing as Angular filters (but can transform in both directions). I'm hesitant to open up the whole global thing. So far, everything is nicely encapsulated and this would break that as well as encourage people to put more code in their views.

tomasbonco:
Ok, but although it's almost not documented (Value Convertors), it looks like it requires a class per filter. Thinking about converting underscore / lodash (huge number of possible filters) into such a system is unreal. (Maybe I'm wrong and it's easier)

It might encourage people to put more code in their views, but they will do it anyway plus they have to "hack" view-models like this (still "elegant" way):
```
this.window = window;
```

I believe window and document are super globals so they should be accessible from everywhere.

EisenbergEffect:
That's a good argument. Let me think about it a bit.

EisenbergEffect:
I'm not going to accept this...yet. I want to keep this open though. Before we evaluate this, we want to take a look at how we are going to remove $parent, which I think might affect this. Once we resolve that issue, we can come back to this and review.

EisenbergEffect:
I'm closing this for now since there's no way we'll be able to merge the actual PR by the time we get around to addressing it. I am keeping the issue itself open so we can revisit it again later.

PR #18, opened 2015-02-07
feat(map-observation): implement map observation

martingust:
Basically I copied the array observe structure. Change records use key instead of index, and we don't have to deal with splices, which simplifies stuff.


EisenbergEffect:
Ok. I went ahead and merged this. We're going to do some refactoring on it in a bit. Let me get the other stuff in too and then we can talk about making some improvements.

Since you are acquainted with this code, I'd like for you to think about maybe doing more in depth work on the repeat behavior.

PR #24, opened 2015-02-11
fix(observer-locator) ObserverLocator assumes OoPropertyObserver when property is undefined

### LINKED ISSUE #23: ObserverLocator assumes OoPropertyObserver when getObserver is called for an undefined property
When getObserver is called for an undefined property an instance of OoPropertyObserver is returned.

When the property later becomes defined and is not a primitive property (has a getter/setter), the incorrect observer type has already been returned to the templating system so property change events cannot be observed.

I'm thinking we should return an instance of a new class: `UndefinedPropertyObserver`.  This new class would listen for the property to become defined using Object.observe.  When this happens it would do the following:

1. call ObserverLocator.getObserver to retrieve the correct "actual" observer.
2. delegate the subscribe, getValue and setValue methods to the "actual" observer.
3. funnel any existing callbacks into the "actual" observer via it's subscribe method.

*note:*   `UndefinedPropertyObserver`'s setValue method would need to have special handling to ensure the appropriate observer is created and change events are published.

**Edit**
specifically talking about cases where Object.getOwnPropertyDescriptor returns `undefined`.  (As opposed to cases where it returns `{value: undefined, writable: true, enumerable: true, configurable: true}`)

jdanyow:
Added UndefinedPropertyObserver, to be used when the property does not exist.

Fixes #23

### DIFF
--- a/src/observer-locator.js
+++ b/src/observer-locator.js
@@ -69,8 +69,8 @@ function createObserversLookup(obj) {
   return value;
 }
 
-function createObserverLookup(obj) {
-  var value = new OoObjectObserver(obj);
+function createObserverLookup(obj, observerLocator) {
+  var value = new OoObjectObserver(obj, observerLocator);
 
   try{
     Object.defineProperty(obj, "__observer__", {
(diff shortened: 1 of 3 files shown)


jdanyow:
Tried to be surgical with the changes.  Ended up fitting pretty nicely into the original design although I'm sure all this stuff is going to get refactored pretty heavily given the upcoming perf changes and all the other considerations people have thrown at this area lately.

PR #28, opened 2015-02-12
fix(Binding) support data-* attributes

jdanyow:
Fixes aurelia/templating/#17

### DIFF
--- a/src/binding-expression.js
+++ b/src/binding-expression.js
@@ -28,6 +28,10 @@ class Binding {
   constructor(observerLocator, sourceExpression, target, targetProperty, mode, valueConverterLookupFunction){
     this.observerLocator = observerLocator;
     this.sourceExpression = sourceExpression;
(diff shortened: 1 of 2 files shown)


jdanyow:
Took a shot at fixing this although I'm not sure this is the correct/intended solution.

EisenbergEffect:
That might work. Another thing to consider is if this should be built into the element observer itself. I'm not sure. While pondering this we need to consider another scenario: aria- attributes which I don't believe have a property equivalent. So, in this case we must do attribute updates.

We need to do some work on ElementObserver anyway...I'm wondering if we can make it smart enough to handle both of these scenarios. What do you think?

jdanyow:
That makes a lot more sense.  Putting the fix in the ElementObserver keeps this element related business in one spot.  Seems so obvious now that you've said it.  I didn't go deep enough.

PR #29, opened 2015-02-13
fix(observer-locator) support data-* and aria-* attributes

jdanyow:
Fixes aurelia/templating/#17


jdanyow:
attempt 2:
* EventManager.getElementHandler gets a second argument, the property name.  This way the ObserverLocator knows precisely when to use an ElementObserver (when obj is an input kind of element and property is a value prop).

* ElementObserver renamed to ValueAttributeObserver

* New observer: DataAttributeObserver.  Used when obj is an Element and property is a data-* or aria-* attribute.  Throws on subscribe.

PR #40, opened 2015-02-22
fix(ElementObserver) support namespaced attributes

### LINKED ISSUE #34: Binding to Namespaced Attributes
When bind to attributes(one-way) there are a couple of options you can either bind with attr-name.bind="variable" (also tried one-way and one-time) or using interpolation attr-name="${variable}", either way though if you try to bind to a namespaced element such as xlink:href you currently get:

```bash
Uncaught NamespaceError: Failed to execute 'setAttributeNS' on 'Element': '' is an invalid namespace for attributes.
```

For have the following in a controller page.js:

```javascript
export class page {
	constructor(){
		this.icon = 'blah';
	}
}
```
and the following in page.html:
```html
<template>
  <svg class="icon ${icon}">
    <use xlink:href="${icon}"></use>
  </svg>
</template>
```

As I said either of the bind's above is throwing the given error. It appears that the bug is in property-observation, in that the observer is either setting to a raw attribute or a null namespaced attribute which is causing the error.

jdanyow:
Fixes #34

### DIFF
--- a/src/observer-locator.js
+++ b/src/observer-locator.js
@@ -7,8 +7,7 @@ import {
   SetterObserver,
   OoObjectObserver,
   OoPropertyObserver,
-  ValueAttributeObserver,
-  DataAttributeObserver
+  ElementObserver
 } from './property-observation';
 import {All} from 'aurelia-dependency-injection';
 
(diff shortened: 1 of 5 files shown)


jdanyow:
Reverted some changes from my earlier pull requests that went away from ElementObserver in favor of observers for specific property scenarios.

Added a bunch more tests.

PR #41, opened 2015-02-22
feat(ObserverLocator) observe computed properties via their dependencies

### LINKED ISSUE #15: observing computed properties via their dependencies
@Alxandr and I were brainstorming how to get around using dirty-checking when binding computed properties (properties that have a getter function).

One idea we were throwing around was to provide a way declare the computed property's dependencies.  This way aurelia could observe the computed property's dependencies and recompute the property value only when the dependencies change.  If this isn't something that should be put in Aurelia directly an adapter could provide this logic (https://github.com/aurelia/binding/pull/10).

Thoughts?

----------------------------------

**First attempt (no worky):**
```js
var person = { firstName: 'Don', lastName: 'Draper' };

Object.defineProperty(person, 'fullName', {
  enumerable: true,
  configurable: true,
  get: function() { return this.firstName + ' ' + this.lastName; },
  dependencies: ['firstName', 'lastName']  // wondering if getOwnPropertyDescriptor will return this...
});


var descriptor = Object.getOwnPropertyDescriptor(person, 'fullName');
descriptor.dependencies  // undefined :(  bummer....
```

**Second Attempt (works but not ES6 style/if-stmt in getter is bad idea)** 
```js
var person = { firstName: 'Don', lastName: 'Draper' };

Object.defineProperty(person, 'fullName', {
  enumerable: true,
  configurable: true,
  get: function(returnDependencies) {
    if (returnDependencies)
      return ['firstName', 'lastName'];    
    return this.firstName + ' ' + this.lastName; 
  }
});


var descriptor = Object.getOwnPropertyDescriptor(person, 'fullName');
descriptor.get(true);
```

**Third Attempt (works but ugly/even worse)**
```js
var person = { firstName: 'Don', lastName: 'Draper' };
function fullName() {
  return this.firstName + ' ' + this.lastName; 
}

fullName.dependencies = ['firstName', 'lastName'];

Object.defineProperty(person, 'fullName', {
  enumerable: true,
  configurable: true,
  get: fullName
});

var descriptor = Object.getOwnPropertyDescriptor(person, 'fullName');

descriptor.get.dependencies;
// ["firstName", "lastName"]
```

**Last Attempt**
```js
function declarePropertyDependencies(ctor, propertyName, dependencies) {
  var descriptor = Object.getOwnPropertyDescriptor(ctor.prototype, propertyName);
  descriptor.get.dependencies = dependencies;
}

export class Person {
  constructor() {
    this.firstName = 'Don';
    this.lastName = 'Draper'; 
  }

  get fullName() {
    return this.firstName + ' ' + this.lastName;
  }
}

declarePropertyDependencies(Person, 'fullName', ['firstName', 'lastName']);
```

**Variation of last attempt, enabling multiple props to be configured in one call**
```js
function markComputedProperties(object, props) {
  var keys = Object.keys(props);
  for (var i = 0, l = keys.length; i < l; ++i) {
    var descriptor = Object.getOwnPropertyDescriptor(object.prototype, key);
    descriptor.get.dependencies = props[key];
  }
}

export class Person {
  constructor() {
    this.firstName = 'Don';
    this.lastName = 'Draper'; 
  }

  get fullName() {
    return this.firstName + ' ' + this.lastName;
  }
}

markComputedProperties(Person, {
  fullName: ['firstName', 'lastName']
});
```

jdanyow:
Fixes #15

### DIFF
--- a/src/computed-observation.js
+++ b/src/computed-observation.js
@@ -1,0 +1,73 @@
+export class ComputedPropertyObserver {
+  constructor(obj, propertyName, descriptor, observerLocator){
+    this.obj = obj;
+    this.propertyName = propertyName;
+    this.descriptor = descriptor;
+    this.observerLocator = observerLocator;
+    this.callbacks = [];
+  }
+
+  getValue(){
+    return this.obj[this.propertyName];
+  }
+
+  setValue(newValue){
+    throw new Error('Computed properties cannot be assigned.');
+  }
+
+  trigger(newValue, oldValue){
+    var callbacks = this.callbacks,
+        i = callbacks.length;
+
+    while(i--) {
+      callbacks[i](newValue, oldValue);
+    }
+  }
+
+  evaluate() {
+    var newValue = this.getValue();
+    if (this.oldValue === newValue)
+      return;
+    this.trigger(newValue, this.oldValue);
+    this.oldValue = newValue;
+  }
+
+  subscribe(callback){
+    var dependencies, i, ii;
+
+    this.callbacks.push(callback);
+
+    if (this.oldValue === undefined) {
+      this.oldValue = this.getValue();
+      this.subscriptions = [];
+
+      dependencies = this.descriptor.get.dependencies;
+      for (i = 0, ii = dependencies.length; i < ii; i++) {
+        // todo:  consider throwing when a dependency's observer is an instance of DirtyCheckProperty.
+        this.subscriptions.push(this.observerLocator.getObserver(this.obj, dependencies[i]).subscribe(() => this.evaluate()));
+      }
+    }
+
+    return () => {
+      this.callbacks.splice(this.callbacks.indexOf(callback), 1);
+      if (this.callbacks.length > 0)
+        return;
+      while(this.subscriptions.length) {
+        this.subscriptions.pop()();
+      }
+      this.oldValue = undefined;
+    };
+  }
+}
+
+export function hasDeclaredDependencies(descriptor) {
+  return descriptor && descriptor.get && !descriptor.set
+    && descriptor.get.dependencies && descriptor.get.dependencies.length;
+}
+
+export function declarePropertyDependencies(ctor, propertyName, dependencies) {
+  var descriptor = Object.getOwnPropertyDescriptor(ctor.prototype, propertyName);
+  if (descriptor.set)
+    throw new Error('The property cannot have a setter function.');
+  descriptor.get.dependencies = dependencies;
+}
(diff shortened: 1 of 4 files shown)


jdanyow:
Implemented this as a system-supplied adapter.  To use in the skeleton, switch to aurelia-main and add the following:

**main.js**
``` javascript
import {ComputedObservationAdapter, ObjectObservationAdapter} from 'aurelia-framework';
...
...
aurelia.container.registerSingleton(ObjectObservationAdapter, ComputedObservationAdapter);
```

Then to avoid dirty checking Welcome.fullName, make the following changes:

**welcome.js**
``` javascript
import {declarePropertyDependencies} from 'aurelia-framework';

export class Welcome{
  ...
  ...
}

declarePropertyDependencies(Welcome, 'fullName', ['firstName', 'lastName']);
```

I'm guessing we'll take care of the main,js piece for people by adding something to the bootstrapper?

Other option was to hard code this thing in the ObserverLocator.  I wasn't sure about doing that.  Seems like there's more work to do in this area- need to come up with a design for handling properties with setters.  Something similar to Object.getNotifier maybe.

EisenbergEffect:
I'm thinking we should just go ahead and "hard code" this one. What do you think? I don't think someone should have to turn this on. I think it's a reasonable default behavior.

FYI Once we get ES7 decorator support on Babel we will be able to do this:

```javascript
export class Welcome {
  @dependencies('firstName', 'lastName')
  fullName(){
    return this.firstName + ' ' + this.lastName;
  }
}
```

Thoughts on all this?

jdanyow:
That's awesome- is [this](https://github.com/Microsoft/TypeScript/issues/1557) the best link for the ES7 decorators?  I'm trying to figure out how closely the way we're currently "decorating" the property aligns with the future implementation.  My guess is it's not close at all.

I agree the observe-via-dependencies option should be available by default- I'll refactor the pull request to discontinue using adapters and hard-code the condition in the ObserverLocator.

EisenbergEffect:
@jdanyow That's not the latest information on decorators. There's design work on that happening elsewhere. I'm not sure I should post the link here because I'm not sure if the repo author is prepared to handle community questions at this point. I'll put it in the private gitter. Suffice it to say, the design of decorators allows us to write code like above which can be used to attach metadata to the getter function.

jdanyow:
made the changes- now the process would be this:

**welcome.js**
``` javascript
import {declarePropertyDependencies} from 'aurelia-framework';  // <-- add this

export class Welcome{
  ...
  ...
}

declarePropertyDependencies(Welcome, 'fullName', ['firstName', 'lastName']);  // <-- and this

PR #48, opened 2015-03-09
feat(ElementObserver) style binding

### LINKED ISSUE #38: support style binding
@EisenbergEffect Thoughts on what the binding expressions would look like?

[Knockout style](http://knockoutjs.com/documentation/style-binding.html) ?

Or what if the framework had attached behaviors built from the first column of this list (maybe prefixed with `css-` or `style-`):

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Properties_Reference

Assign to me if you want.

jdanyow:
Fixes #38

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "karma": "^0.12.28",
     "karma-babel-preprocessor": "^4.0.0",
     "karma-chrome-launcher": "^0.1.7",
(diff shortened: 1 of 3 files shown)


jdanyow:
@EisenbergEffect  this one is ready too- provided you like the approach- there was a lot of discussion around this one...

PR #55, opened 2015-03-28
fix(SelectValueObserver): handle late bound option values

### LINKED ISSUE #54: select-value binding doesn't update when static options are bound
related to https://github.com/aurelia/templating-resources/issues/21

In this case the options are static but each option's value/model are data bound.  

view-model:
``` javascript
this.value = true;
```

view:
``` html
  <select value.bind="value">
    <option model.bind="null">null</option>
    <option model.bind="false">false</option>
    <option model.bind="true">true</option>
  </select>
```

The select-value binding happens before the model bindings on the options.  Since these bindings don't mutate the dom, the mutation-observer used in https://github.com/aurelia/templating-resources/issues/21 doesn't help.

Writing tests for the fix... will submit PR.

jdanyow:
Fixes #54

### DIFF
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -382,6 +382,11 @@ export class SelectValueObserver {
     // assign and sync element.
     this.value = newValue;
     this.synchronizeOptions();
+    // queue up an initial sync after the bindings have been evaluated.
+    if (this.element.options.length > 0 && !this.initialSync) {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
All good to go? or were you still planning some additional features?

jdanyow:
this one is ready

PR #76, opened 2015-04-23
feat(AccessKeyed): fully observe access-keyed expressions

### LINKED ISSUE #64: Assigning an Array value by index does not notify subscribers.
**viewModel**
`array = ['foo']`

**view**
`<div>${array[0]}</div>`

The above will output foo, as expected. However, if you index-assign a new value, e.g. `array[0] = 'bar'`, the output is not updated.

### LINKED ISSUE #75: AccessKeyed should observe the object[key]
Currently AccessKeyed sets up an observer for the property that accesses the object instance (scope.object) and the property that accesses the key value (scope.key) but it does not observe the object's property identified by the key.  

The current behavior makes sense when the object is an array instance (although [it's been argued](https://github.com/aurelia/binding/issues/64) that Aurelia should use dirty-checking in cases where AccessKeyed accesses an array).

[Plunker demonstrating the issue](http://plnkr.co/edit/DIRJiaSzUwGbmykWybs5?p=preview) (thanks @mattduffield)

jdanyow:
Fixes #75, Fixes #64, Fixes aurelia/templating-binding#26

### DIFF
--- a/src/access-keyed-observer.js
+++ b/src/access-keyed-observer.js
@@ -1,0 +1,70 @@
+export class AccessKeyedObserver {
+  constructor(objectInfo, keyInfo, observerLocator, evaluate) {
+    this.objectInfo = objectInfo;
+    this.keyInfo = keyInfo;
+    this.evaluate = evaluate;
+    this.observerLocator = observerLocator;
+
+    if (keyInfo.observer) {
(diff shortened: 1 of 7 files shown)


jdanyow:
Lots of tests to add...  `AccessKeyedObserver` is a specialized version of `CompositeObserver` that knows how to manage changes to the object value or the key value.

jdanyow:
Added tests, tested in karma chrome runner and IE runner.  This is ready.

PR #82, opened 2015-04-28
feat(CallExpression): add $event to scope, use call args

### LINKED ISSUE #46: support for evaluating behavior scoped arguments
In app-contacts you can do
```
<a href="#" click.delegate="$parent.select(contact)">
</a>
```
where you can pass view scoped arguments like `contact`

If I have a custom behavior I can't do the same
```
<div oribella-tap="end.call: onTap(e,data,contact);"></div>
```
where `e` and `data`are behavior scoped and `contact` view scoped. It would be nice if the `CallExpression` could support evaluating behavior scoped properties somehow.
Let me know if you need more info.

jdanyow:
Fixes #46

### DIFF
--- a/src/call-expression.js
+++ b/src/call-expression.js
@@ -35,8 +35,12 @@ class Call {
     }
 
     this.source = source;
-    this.targetProperty.setValue((...rest) => {
-      return this.sourceExpression.evaluate(source, this.valueConverterLookupFunction, rest);
+    this.targetProperty.setValue($event => {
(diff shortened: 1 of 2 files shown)


jdanyow:
@EisenbergEffect is this related:  https://github.com/aurelia/templating-binding/issues/9 ?  Do we need the equivalent of knockout's `$data` added to the scope in the same fashion as `$event` ?

Issue #80, opened 2015-04-26
The value of a model property that is bound to the 'class' attribute of an 'svg' tag is instead ending up in a 'className' attribute 


ifyio:
Given the following custom element template...

```
<!-- svg.class="search" -->
<template>
  <svg role="img" class="${svg.class}">
    ...
  </svg>
</template>
```

The rendered output is as follows...

```
<svg role="img" class="${svg.class} au-target" className="search">
    ....
</svg>
```

Notice that `${svg.class}` is still present in the `class` attribute of the `svg` tag and the values that should have replaced `${svg.class}` have instead been added to a separate attribute called `className`. The expected behaviour is that the value of `svg.class` should be added to the class attribute of the `svg` tag.

jdanyow:
thanks @ifyio - will look into this

jdanyow:
http://stackoverflow.com/a/29525743/725866

Aurelia [remaps](https://github.com/aurelia/templating-binding/blob/master/src/binding-language.js#L19) the "class" attribute name to the dom api property name: "className".

In SVG elements this doesn't work properly because `className` is not a string, but an instance of `SVGAnimatedString`.

jdanyow:
related to aurelia/binding#33

Issue #33, opened 2015-02-14
class.bind currently overrides any existing class definitions on the element


shyamseshadri:
If i have

```html
<div class="todo-card" class.bind="todo.completed ? 'completed' : ''"></div>
```
I expect that the div either becomes
```html
<div class="todo-card completed">
```
or
```html
<div class="todo-card">
```
Currently it becomes
```html
<div class="completed">
```
or
```html
<div class="">
```
I'm not sure this would be the expected behavior. I would not want to do 
```html
<div class="todo-card ${todo.completed ? 'completed': ''}"> 
```
in a lot of cases

paullryan:
@shyamseshadri something seems to have go wrong with your formatting as nothing is showing after your statements for code (e.g. "if I have" is followed by blank).

EisenbergEffect:
`class.bind` is expected to overwrite the existing classes. If you do not want to do this, you will want to use interpolation instead. Something like this `class="some-class another-class ${boundClass}"`

olanod:
I'd like to reopen this, there should be a way to do this class interpolation thing without overwriting current classes but without using the example of @EisenbergEffect.  
I have an use case where I want to add a dynamic class to a custom element. `<my-element class="${myClass}"></my-element>` but the element in its inner logic toggles a class from Js which I don't(and shouldn't) control in this scope.  
I don't know there should be a special syntax for this or a different behavior but I think it should be possible.

olanod:
Not sure if it's important but also note that `class.bind="..."` and `class="${...}"` overwriting deletes the `au-target` class also.

EisenbergEffect:
With respect to `au-target` it's no longer important to the framework by the time that the binding expression is applied.

EisenbergEffect:
@jdanyow Can you an @olanod continue discussion here on his use case?

olanod:
So as I previously mentioned. I want to set a class conditionally
```
<my-element class="${condition?'active':''}"></my-element>
```
but this will replace the classes the element already has. In my case the classes I want to keep are added dynamically from inside the custom element(or from anywhere anytime?).
E.g.
In the custom element logic in the `bind` method I add a class with `element.classList.add('fuu')`, it gets added to the host element but then sometime in the future some event changes the `condition` and the element gets the `'active'` class but it also deletes any previous classes it could have.

shaunluttin:
How was this issue resolved?

atsu85:
I haven't seen that issue, but you could use string interpolation instead of .bind

PR #95, opened 2015-05-29
feat(ClassObserver): enable multiple class bindings

jdanyow:
fixes aurelia/templating-resources#62, #80, #33

### DIFF
--- a/src/class-observer.js
+++ b/src/class-observer.js
@@ -1,0 +1,53 @@
+export class ClassObserver {
+  constructor(element) {
+    this.element = element;
+    this.doNotCache = true;
+    this.value = '';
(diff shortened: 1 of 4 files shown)


jdanyow:
https://github.com/aurelia/templating-binding/pull/42 goes with this

jdanyow:
@EisenbergEffect I've switched from `instanceof` to the `doNotCache` property.  Also streamlined `ClassObserver.setValue` a bit to avoid some unnecessary looping on the initial call and to avoid re-allocating the dictionary object on every call.

EisenbergEffect:
@jdanyow You are killing it. Awesome!

PR #96, opened 2015-05-29
fix(AST): do not coerce operands of || or && to booleans

jdanyow:
fixes aurelia/templating-binding#37


jdanyow:
tested the reported use case as well as with the `show` and `if` binding commands.

PR #107, opened 2015-06-09
fix(OoObjectObserver): regression issue with String.length observation

### LINKED ISSUE #106: Can't bind to string.length
Binding to string.length uses Oo in the new release which throws:

```
TypeError: Object.observe cannot observe non-object
    at Function.observe (native)
    at OoObjectObserver.subscribe (http://localhost:82/jspm_packages/github/aurelia/binding@0.7.1/property-observation.js:142:20)
    at PathObserver.updateRight (http://localhost:82/jspm_packages/github/aurelia/binding@0.7.1/path-observer.js:41:40)
    at new PathObserver (http://localhost:82/jspm_packages/github/aurelia/binding@0.7.1/path-observer.js:25:16)
    at AccessMember.connect (http://localhost:82/jspm_packages/github/aurelia/binding@0.7.1/ast.js:415:24)
    at Binary.connect (http://localhost:82/jspm_packages/github/aurelia/binding@0.7.1/ast.js:760:36)
    at Binding.bind (http://localhost:82/jspm_packages/github/aurelia/binding@0.7.1/binding-expression.js:70:42)
    at BehaviorInstance.bind (http://localhost:82/jspm_packages/github/aurelia/templating@0.12.1/behavior-instance.js:60:23)
    at View.bind (http://localhost:82/jspm_packages/github/aurelia/templating@0.12.1/view.js:68:26)
    at BehaviorInstance.bind (http://localhost:82/jspm_packages/github/aurelia/templating@0.12.1/behavior-instance.js:72:23)
```

In the previous release this seemed to work ok

My binding example:

```
<span if.bind="badge.length > 0" class="label label-default">${badge}</span>
```

I can workaround with 

```
if.bind="badge"
``` 

For now, but someone might actually be interested in binding to length and displaying it somewhere

jdanyow:
Fixes #106

### DIFF
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -114,7 +114,9 @@ export class OoObjectObserver {
 
     if (this.callbackCount === 0) {
       this.handler = this.handleChanges.bind(this);
-      Object.observe(this.obj, this.handler, ['update', 'add']);
+      try {
+        Object.observe(this.obj, this.handler, ['update', 'add']);
+      } catch(_) {}
     }
 
     this.callbackCount++;
@@ -137,7 +139,9 @@ export class OoObjectObserver {
 
     this.callbackCount--;
     if (this.callbackCount === 0) {
-      Object.unobserve(this.obj, this.handler);
+      try {
+        Object.unobserve(this.obj, this.handler);
+      } catch(_) {}
     }
   }
 


EisenbergEffect:
Does this allow string.length to be observed? What was the behavior before?

jdanyow:
This restores the behavior from before.  I mistakenly removed the try-catch.  Now I know why it's there...

I don't think we need to do anything to observe string length- if the length changes the property value must have changes so the path observer will take care of this.  Maybe we can add some special handling in the ObserverLocator for this scenario to improve perf and eliminate the try-catch that is found in ObserverLocator and the OoObjectObserver.  Will need to research.

EisenbergEffect:
Ok. It's coming back to me now. I think I remember writing this ;) I'll get this out momentarily. We can find a way to improve it later. For now, on to benchmarking :)

PR #110, opened 2015-06-10
fix(ClassObserver): handle null and undefined

### LINKED ISSUE #109: Null no longer works in class.bind
Not sure if this should be considered a bug or a breaking change in the latest releases of the binding library, but since I updated to the latest version the ClassObserver.setValue method crashed telling me split is not available on undefined.

After an hour or so of debugging I found out that I can no longer use null values in class.bind bindings. For exemple in my VM I had a property defined as sortCss = null (that I in some cases set to something else) which now crashed. When I changed it to sortCss = "" my application works again.

Not really an issue now when I know I have to set it to an empty string (or leave it undefined?), but...

jdanyow:
Fixes #109

### DIFF
--- a/src/class-observer.js
+++ b/src/class-observer.js
@@ -16,15 +16,17 @@ export class ClassObserver {
         names, name, i;
 
     // Add the classes, tracking the version at which they were added.
-    names = newValue.split(' ');
-    i = names.length;
-    while(i--) {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Releasing coming in a few minutes.

PR #118, opened 2015-06-19
fix(Conditional): short circuit evaluation of left/right

### LINKED ISSUE #116: Calling member function of nullable
*  `${selectedItem.date.toISOString()}`
*  `${selectedItem.date && selectedItem.date.toISOString()}`
*  `${selectedItem.date ? selectedItem.date.toISOString(): ''}`
*  `${selectedItem.date != null ? selectedItem.date.toISOString(): ''}`

Either version throws an error in console, when selectedItem.date is null.
Tested FF, Chrome, IE11.

Demo: http://plnkr.co/edit/sHgRZxH9Z0KODactz1JX?p=preview

jdanyow:
Fixes #116

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -1,6 +1,7 @@
 import {PathObserver} from './path-observer';
 import {CompositeObserver} from './composite-observer';
 import {AccessKeyedObserver} from './access-keyed-observer';
+import {ConditionalObserver} from './conditional-observer';
 
 export class Expression {
   constructor(){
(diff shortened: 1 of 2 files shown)


jdanyow:
@EisenbergEffect  - Initially I thought we could fix this with two changes:

1. Change the `Conditional` expression class to lazily `connect` the true/false sides of the conditional
2. Replace the `CompositeObserver` with a specialized `ConditionalObserver` that unsubscribes from the true/false side of the conditional depending on what the condition value.

The current state of the PR includes these changes and partially fixes the issue. 

The remaining issue is unsubscribing from the true/false side of the conditional expression doesn't work when the true/false expression uses a composite observer such as `CompositeObserver` / `PathObserver` / `AccessKeyedObserver` / `ConditionalObserver`.  These observers don't unsubscribe their inner observers when unsubscribe is called.  The inner observers are not unsubscribed until the composite observer instance is disposed.  To complete this fix I think we'll need to update the composite observer classes to unsubscribe from their inner observers when unsubscribe is called and re-subscribe when subscribe is called.

Do you think I'm on the right track?  I was thinking this would be something we need to fix ahead of the perf stuff since it might influence that work in some ways.

EisenbergEffect:
All those observers are going to go away with the binding engine optimization. So, I would hate to spend time on all that now.

jdanyow:
k- closing this then

PR #142, opened 2015-07-30
fix(handleDelegateEvent): stop propagation beyond DOM boundary

jdanyow:


EisenbergEffect:
Fixed  and released ;)

stoffeastrom:
What does this mean? This could potentially be a problem for others plugins. Default the DOM boundary seems to be `document` which hopefully will not be a problem. Why do you need to call `event.stopPropagation()`?

stoffeastrom:
@jdanyow @EisenbergEffect any comments regarding the `event.stopPropagation` call? Why was it needed. It would be better if this kind of calls could be avoided but unfortunately it sometimes is needed just wondering if this could be solved differently.

EisenbergEffect:
The idea behind the delegated event is that the actual event handler is registered somewhere higher in the tree, such that the event bubbles up to that higher element and it is handled there. The problem is that our earlier version broke component encapsulation by always registering the handlers on the document, allowing internal events to bubble out of components. But it also didn't work with the shadow dom, which prevented events from bubbling out of the shadow root. The new implementation uses component boundaries as the location to register the actual handlers, which allows both fixes the shadow dom bug and allows our native components to function more like standard web components. Thecall to `stopPropagation` in this case was added to solve an IE bug. Though it does make sense that propagation should stop.

@stoffeastrom Is there a bug that this is causing for you?

stoffeastrom:
Not at the moment. I was just worried since `event.stopPropagation` could cause troubles. For instance my gesture framework only adds event listeners on `document` and doing a `event.stopPropagation()` on any of the native DOM events (example mousedown, mousemove, mouseup) could break the event flow completely.

If it's only for internal events it makes sense. If not, then I'm still worried. As I wrote doing a `event.stopPropagation()` should only be used as a last resort in my opinion.

stoffeastrom:
@EisenbergEffect this is affecting me horrible right now. I just tested and it breaks everything :(

```
<div class="page-host" mousedown.delegate="test();">
    <router-view></router-view>
  </div>
```
Added this and now my sortable plugin stopped working

EisenbergEffect:
Us there some reason you can' t use trigger?

stoffeastrom:
That was an example... I can't control what any user does...

stoffeastrom:
Anyone using the plugin could break it by adding that

EisenbergEffect:
I wonder if what we should do is re-target the events. I think that's what shadow dom does. Idea would be that we stop propagation of the original event. But, then we fire a new event of the same type, but retargeted so that it looks like it came from the component's element rather than the element inside the component. @jdanyow Do you think this would allow IE to function correctly?

stoffeastrom:
I'm not sure I understand what the root cause was for changing this. But reading through your previous answer makes me a bit more concerned. What you are now saying is that the event listeners are now added on the closest component level? This means if you have a custom element in a repeat and register an event on an element in that, a DOM event listener would be added for each item in the repeat. This could lead to severe performance problems.

1. Regarding the component encapsulation. I don't agree. This is handled by the framework and thus doesn't break the encapsulation per see from the user point of view.

2. Regarding the shadow DOM. What was the reason for the events not bubbling out of the shadow root?

3. What was the issue in IE?

stoffeastrom:
@EisenbergEffect I just tested and it adds an event listener for each item in the repeat. Is this really intended?

Looked through the shadow DOM draft and indeed there are some event that are scoped
> 5.1 Events that are not leaked into ancestor trees

>For the following events, scoped flag is initialized to true by default if the events are dispatched by the user agent:

>abort
>error
>select
>change
>load
>reset
>resize
>scroll
>selectstart

Don't know if that was the issue though. 

Also, I don't think we should consider all components as web components. If there is need for some special treatment for the shadow DOM it should be handled with feature detection and not affect other components.

EisenbergEffect:
We will get this fixed this week. I will work on it myself, top priority. I've got some ideas about it, but I first need to do some spec reading and run a few experiements in chrome. I'll report back.

JeroenVinke:
This is also a breaking change for me. click.delegates in anchors don't work after this change. I think I will go back to 0.8.3, because I used .delegate() for everything since it was recommended to do so in the docs

EisenbergEffect:
I have a fix I believe. I hope to get it out on Monday. I will add some
notes here soon.

EisenbergEffect:
The solution I have is pretty simple and involves the following:

* Change the event manager so that it uses some new event properties to find the original event target even in the presence of shadowDOM. The code basically looks like this:
```
function findOriginalEventTarget(event){
  return event.originalTarget || (event.path && event.path[0])
    || (event.deepPath && event.deepPath[0]) || event.target || event.srcElement;
}
```
This code accounts for differences in browsers and differences in shadow dom spec versions. With this code, as long as it's not a *closed* shadow dom, we can retrieve the original event target and thus make event delgation work across shadow dom boundaries.
* With the above fix in place, the adhoc domBoundary property has been removed from elements when it was previously added by the view factory.
* All delegated events can be handled again by the document, thus there is no need to use the dom boundary at all in this process. As a result, the stopping of event propagation has also been removed.
* The dom boundary is only created when there is a real shadow dom or a request for one, but not for components where authors did not specify this. The system does not use this itself, but just makes it available so component authors working in shadow dom can get access to the shadow root.

Merging changes to master shortly. This involves changes in binding and templating. I hope to release this today or tomorrow. Thank you all for your patience and I apologize for the issues.

stoffeastrom:
This sounds great @EisenbergEffect! I think it was really good to catch this now and not later. :+1:

EisenbergEffect:
Release is looking good for tomorrow. Hopefully everyone on this thread will kick the tires on that and let us know if it's all working well for your scenarios.

Issue #152, opened 2015-08-18
text inputs bound to non-existent values show up as "undefined" in composed elements


mbroadst:
relevant plunker: http://plnkr.co/edit/Yzs2RXNJynmkjX3k1RcA

**whats going on here**
This plunker contains an element (MyDialog) which composes another element (MyDialogContent), and binds a single value from the top level all the way to the composed element (object is bound to my-dialog, and my-dialog binds the same object to my-dialog-content's model in `activate`).

**expectations**
  * click "ADD" - all input boxes show up empty
  * click "UPDATE" - input boxes contain the values of relevant properties on bound object

**actual result**
  * click "ADD" - all input boxes show "undefined" (_incorrect_)
  * click "UPDATE" - input boxes contain the values of relevant properties on bound object (_correct_)

Interestingly enough, if you modify the definition of MyDialogContent to contain a `record = {}` at the top, the initial rendered view is correct (each field displays an empty string, or in this case the placeholder text). Once you click "ADD", the broken behavior exhibits itself again.

mbroadst:
This appears to be unrelated to composition as I initially thought, this plunker exhibits the issue with a simple VM: http://plnkr.co/edit/xJoTFhkYCyYtaaGWhSLo

mbroadst:
For further reference (and weirdness), here is a link to a fork of a plunker @jdanyow gave me earlier. I've added three inputs to always bind to the current `editing` member, when you click "Add new planet" the value for `editing.name` shows as undefined, however the other two fields correctly display as empty strings: http://plnkr.co/edit/ulpjUa6EbFpZzgOJmnQ1

jdanyow:
@mbroadst - I took a look at the plunker you put together [here](https://github.com/aurelia/binding/issues/152#issuecomment-132286888).

The issue you're seeing is the inputs are initially blank but when you click add-new they end up with the word `undefined` in them.

What you've uncovered is an inconsistency in the binding code.  When the view is initially bound to the view-model, the binding code does not assign the view-model value to the view when the value is undefined:

https://github.com/aurelia/binding/blob/master/src/binding-expression.js#L83-L85

Later on, when add-new is pressed, this kind of check does not occur and `undefined` IS assigned to the input element's value property, causing the word `undefined` to appear in the input.

We need to decide on what the expected behavior is here, being mindful that the `targetProperty` isn't always an input per-say but could also be a custom element's bindable prop, etc. 

[...]

PR #153, opened 2015-08-19
fix(ValueAttributeObserver): set undefined/null to empty string

mbroadst:
Previously when passed a null or undefined value, a
ValueAttributeObserver would set the value of that html element to
a string representation of that value. This patch changes that
behavior to be more in line with the initial BindingExpression
behavior of setting that value to an empty string.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "karma-coverage": "^0.3.1",
     "karma-ie-launcher": "^0.1.5",
     "karma-jasmine": "^0.3.5",
-    "karma-jspm": "^1.1.5",
+    "karma-jspm": "^2.0.1",
     "object.assign": "^1.0.3",
     "require-dir": "^0.1.0",
     "run-sequence": "^1.0.2",
--- a/src/binding-expression.js
+++ b/src/binding-expression.js
@@ -80,9 +80,7 @@ class Binding {
         });
       }
 
-      if(info.value !== undefined){
-        targetProperty.setValue(info.value);
-      }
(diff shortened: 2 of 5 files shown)


jdanyow:
@mbroadst this PR looks good- nice work!  Minor nit-picky thing:  could you amend the commit message to include a second line that says `Fixes #152`?

Please [sign the CLA](https://docs.google.com/forms/d/1Npd68d32UDUKUYuCDDrW0VN5Lv8p-a6Ziaa2JFKp6Cg/viewform?c=0&w=1) if you haven't yet.

@EisenbergEffect as part of this should we remove this check:

https://github.com/aurelia/binding/blob/master/src/binding-expression.js#L83

EisenbergEffect:
Great! @jdanyow Would you mind sending a PR for the binding expression fix? As you know I've been travelling a lot this week...this evening I should be spending several hours reviewing and processing the latest PRs. So, hopefully I will get it in tonight. After that I think we'll make several patch releases. Those might go out tomorrow, depending on time.

mbroadst:
@jdanyow fixed the commit message

@EisenbergEffect I think he was asking your opinion as to whether we _should_ remove that from BindingExpression. I can include that in this PR if you think it's a good idea

EisenbergEffect:
Yes, if I recall, that undefined check was added very early on before we had the more advanced collection of observer we have now. I think it was added as a solution to this very problem, which is fixed more effectively by the new PR. Feel free to amend your PR with this fix as well.

PR #167, opened 2015-09-10
check old value against undefined on map set

benib:
This checks the `oldValue` against `undefined` to have type `update` if the oldValue was set but falsy.

### DIFF
--- a/src/map-observation.js
+++ b/src/map-observation.js
@@ -18,7 +18,7 @@ class ModifyMapObserver extends ModifyCollectionObserver {
 
     map['set'] = function () {
       var oldValue = map.get(arguments[0]);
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
@jdanyow Can you review this? Thanks!

benib:
Here is a plunkr to show the problem. Click the change button to update the two Maps (dataA and dataB). http://plnkr.co/fDfOx9

The item in dataA map is `0` in the beginning, so we see two values after the change although the update just set a new value for the same key.

dataB shows correct behaviour, as item value was `1` in the beginning.

jdanyow:
looked good- added test and merged- thanks @benib !

PR #171, opened 2015-09-25
fix(ObserverLocator): enable adapter installation after instantiation

jdanyow:
This change also streamlines the ObjectObservationAdapter API, resulting in an easier to implement and more performant interface.

Fixes jdanyow/aurelia-computed#8

### DIFF
--- a/src/observer-locator.js
+++ b/src/observer-locator.js
@@ -18,7 +18,6 @@ import {
   StyleObserver
 } from './element-observation';
 import {ClassObserver} from './class-observer';
-import {All} from 'aurelia-dependency-injection';
 import {
   hasDeclaredDependencies,
   ComputedPropertyObserver
@@ -53,12 +52,12 @@ function createObserverLookup(obj, observerLocator) {
 }
 
 export class ObserverLocator {
-  static inject(){ return [TaskQueue, EventManager, DirtyChecker, All.of(ObjectObservationAdapter)]; }
+  static inject(){ return [TaskQueue, EventManager, DirtyChecker]; }
   constructor(taskQueue, eventManager, dirtyChecker, observationAdapters){
     this.taskQueue = taskQueue;
     this.eventManager = eventManager;
     this.dirtyChecker = dirtyChecker;
-    this.observationAdapters = observationAdapters;
+    this.adapters = [];
   }
 
   getObserver(obj, propertyName){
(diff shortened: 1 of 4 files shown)


jdanyow:
@EisenbergEffect could you review these changes to the adapter API?  

I removed the `All.of(ObjectObservationAdapter)` in favor of a simple `addAdapter(adapter)` method.  I also remove the `handlesProperty` method in favor of a single `getObserver` method on the adapter that returns null when it can't handle the object/property.

I figured while we're making changes to binding that will impact observation adapters we might as well streamline this interface.

I couldn't remove the dep on dependency-injection because it's used in a static method on the BindingExpression class.

EisenbergEffect:
Yep, I'm cool with this. Please go ahead and merge it.
Also, regarding removing DI, I think it would be good to create a static interface for the binding library. I can see lot of uses for it. I'm thinking of maybe export an object `bindingSystem` and then having some APIs on that:

```
let const bindingSystem = {
  getObserver(object, propertyName);
  observe(object, propertyName);
  parseExpression(expr);
  addBindingAdpater(adapter);
};
```

We can flesh that out. But basically, I think we should offer this without having to go through the DI. That would simplify our infrastructure in a number of places, particularly related to testing components. It would also make this usable without a reference to DI and enable using this library easier outside of Aurelia. The framework could still register all the underlying services with the Aurelia DI as well though. So, there could be an api like this on the binding system.

```
let const bindingSystem = {
   registerWithContainer(container) {
    //register all underlying services
  }
};
```

Actually, BindingSystem, should probably be a class that we instantiate and then export a singleton instance of so we can also register it with the container, as well as all the underlying pieces.

The result is much easier to use API for observation, much more like Object.observe. Instead you could have `BindingSystem.observe(myObject, 'someProp', callback)` for example.

Ok, this was just thinking out loud so sorry if this isn't all clear. We can discuss it more when you both can sync up. Think about it and let me know what you think.

EisenbergEffect:
Ok, here's something a bit clearer. We can create something like this:

```
export class BindingSystem {
  static instance; //singleton instance

  constructor(parser, observerLocator,...all binding services here...) {}

  //static api
  static observe(obj, propName, callback) {
  let observer = BindingSystem.instance.observerLocator.locate(obj, propName);
  observer.subscribe(callback);
  }
}

BindingSystem.instance = new BindingSystem(...create services...);
```

The framework would also initialize the binding system in the DI container so it's all available both ways. Does this make sense?

jdanyow:
This is nice!  Will be useful both externally in user code and unit tests as well as internally in the templating and binding tests.

EisenbergEffect:
Yep. After you merge these new changes, maybe we could add this. I can review your PR and give you feedback on the final API.

PR #179, opened 2015-10-02
Added getValue for array and collection observer

fragsalat:


jdanyow:
Hi @fragsalat - thanks for your contribution!  Some feedback:

We need to tweak these changes a bit or reconsider adding methods to the collection observation interface.  Slicing the array causes a couple of issues.  First, it means `getValue` doesn't return the value, it returns a new array instance with the same elements.  Repeated calls will allocate new arrays.  The other problem with slice is the ModifyCollectionObserver is used for Arrays, Maps and soon Sets and other types of collections in the future.  These collections don't have a slice method.

A contribution like this that creates a cross-repo (validation->binding) dependency on an internal API like the observers really needs tests.  You'll need to add tests- especially in he validation repo to ensure a future refactoring of the binding code that breaks aurelia-validation is caught before it's released.

Let me run this change by @EisenbergEffect and the validation gurus @PWKad / @janvanderhaegen ) before we do any further work on it.

fragsalat:
@jdanyow Thx for feedback. I needed to return a the array with a slice because otherwise it would give back the array with all references. So in the validation was the case that latestValue was everytime equal to the newValue. If you find a better solution I would be happy If you can show me the commit later on. I also want to apologize for this bad code but it's hard to see and understand all side effects within this short time. Next time I gonna check it better.

PS: Could you tell me how your development setup is? I had problems to test the compiled code because the framework package got it's different dependencies. Do you change these dependencies manually or is there a trick for jspm to point a package with name xy to a specific target?

PWKad:
Sorry just got around to following up on this - I think the problem description is valid (on the Validation side at least) so going through now to understand the changes more in-depth.  Hope to have better feedback tomorrow.

EisenbergEffect:
What was the purpose of this change exactly? I feel like I'm missing some important context when reviewing this.

EisenbergEffect:
Closing this issue for now. We will wait to see if it's necessary after the fixes are made to aurelia-validation to take proper advantage of the binding engine's exposure through created(view).

PR #180, opened 2015-10-06
feat(BindingSystem): add binding system API

jdanyow:
@EisenbergEffect this is a work in progress.  Let me know what you think.

The current design would require us to add the following somewhere in aurelia framework:

```js
import {initializeBindingSystem} from 'aurelia-binding';

initializeBindingSystem(container);
```

To use this outside of an Aurelia application, `initializeBindingSystem()` would be required although we could refactor the BindingSystem class to make the static methods lazy initialize when they're called to support the external or unit-test scenarios more cleanly.

My thinking was this would preserve our ability to have one instance of services like the TaskQueue and EventManager in the container while still enabling external and unit-testing scenarios.

Remaining work:
* remove DI dependency in package.json
* fixup templating where there is a dep on the old BindingExpression.create API

### DIFF
--- a/src/binding-system.js
+++ b/src/binding-system.js
@@ -1,0 +1,87 @@
+import {TaskQueue} from 'aurelia-task-queue';
+import {bindingMode} from './binding-mode';
+import {DirtyChecker} from './dirty-checking';
+import {EventManager} from './event-manager';
+import {ObserverLocator} from './observer-locator';
+import {Parser} from './parser';
+import {BindingExpression} from './binding-expression';
+
+let taskQueue;
+let eventManager;
+let dirtyChecker;
+let observerLocator;
+let parser;
+
+export let __initialized = false;
+
+function initialize(container = null) {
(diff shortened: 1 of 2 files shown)


jdanyow:
related to https://github.com/aurelia/binding/pull/171

jdanyow:
@EisenbergEffect how does it look now?  Any closer?  The method names are verbose.  My intent was to clear up the confusion around observing property-changes vs observing collection-mutation.

Still need to tidy this up and sync up the unit-tests.

jdanyow:
Refactored and fixed up the tests- ready for merge if it looks good.  What do you think of the method names?  Too verbose?

EisenbergEffect:
Looks good except one thing. I think the `assertIntialize` function has a bug. It's trying to call `initialize` which doesn't exist in the module scope...unless I'm seeing something incorrectly. Once that is fixed, please merge.

jdanyow:
good catch- added tests for that as well.

PR #184, opened 2015-10-09
refactor(AST): reduce object allocations during bind

jdanyow:
@EisenbergEffect this is a work in progress but wanted to get your feedback.  All tests passing.  Haven't done any benchmarking yet.  Need to make similar changes to the the InterpolationBinding.


nomack84:
I'm not qualified to asses this PR, but what I know that every time  you do something with less LOC is certainly an improvement...  Nice

EisenbergEffect:
@jdanyow Looks pretty good. I only did a high level review...but seems very promising. Should reduce memory pressure and source code size, which are both great. When do you think you will have time to do the benchmarking. If you can do it by the end of the weekend, and things look good, we can get it into this release. Let me know.

PR #193, opened 2015-10-20
refactor(BindingEngine): use class instead of static object

jdanyow:
@EisenbergEffect how does this look?


EisenbergEffect:
Design-wise it looks good. I think you will have some runtime errors though. In the propertyObserver and collectionObserver methods the `this` variable you are using to access the core services will refer to the subscription object, rather than the BindingEngine instance. So, you will need to capture the outer `this` in those cases.

Issue #194, opened 2015-10-22
API design: bindingEngine


jods4:
This is a bit of follow-up on #158

(1) It was closed without a clear comment, so my guess is that paths can be observed with `expressionObserver`.

Given that this is a superset of `propertyObserver`, do we need to keep both? Why can't we reduce the API surface to just `observer(context, expression)` and `collectionObserver(collection)`? If this is an optimization trick, `observer` could detect that it got a single property and use an optimized codepath. Also, per the discussion in #158 these API are meant to be easy to use, not provide the absolute performance (for this there is `ObserverLocator`).

(2) Another point that bothers me is that we don't have an expression observer for collections. Say: `collectionObserver(context, 'selectedCustomer.orders')`. Currently if I want to observe a list of items and the collection itself may change (e.g. another `selectedCustomer`) than I have to do all the dirty observation / unsubscribe work myself. This is kind of similar to the discussion in #183. I think Aurelia needs a primitive for that. Things like that are trivial in Knockout.

(3) We probably need a syntax to say we want to be notified when array elements change. Kinda like `collectionObserver` but from a string. Here's one example, consider this property on a ViewModel:
```typescript
@computedFrom("selectedCustomer.orders[]")
get ordersCount() { return this.selectedCustomer.orders.length; }
```
This expression changes whenever `selectedCustomer` or `orders` changes, OR when elements are added or removed from `orders`. Of course, this was kind of a bad example because you could observe `selectedCustomer.orders.length` directly and even put this in the View.

But let's consider a more complex example:
```typescript
@computedFrom("selectedCustomer.orders[].price")
get total() { return this.selectedCustomer.orders.reduce((total, o) => total + o.price); }
```
Yes, I even made up a more complicated syntax where you can observe properties on every element of the array, including when elements are added or removed.
Such a scenario would just work in KO.

Doing this today is cumbersome. You need to observe the path `selectedCustomer.orders` yourself, then observe the current array (unobserve when it changes, etc.). Then for each element observe `price`. And when any of this change, redo the computation and store it into a field `total` that you can bind to from the view. This is _a lot_ of plumbing and subscriptions management for something that ought to be simple. 

What do you think?

EisenbergEffect:
[...]

PR #203, opened 2015-10-26
Modified binding-engine API

jods4:
This is related to #194.

I did a few modifications to the `bindingEngine` public API but there are quite a few things that you should be aware of.

1. It's not well tested <_<

2. I considered `bindingEngine` is for "outside" usage. If something depends on it I did not update it.

The main changes are: 

3. I merged `propertyObserver` and `expressionObserver` into a single method `observer(scope, expression)`. It has a fast-path where `expression` is a simple property then it would use simpler code (basically the old `propertyObserver` code).

4. I extended `collectionObserver` with an optional `expression` string. Idea is that you can call it like before: `collectionObserver(myArray)` or you can observe an expression that evaluates to a collection, similar to how `observer()` can observe an expression: `collectionObserver(scope, 'selectedUser.orders')`, which is handy.

Some implementation notes:

5. For my collection expression observer, I had to add an `evaluate` method to the result of `observer(scope, expr)`. It is handy and evaluates `expr` in the context of `scope`. I did not publicly publish it into the `PropertyObserver` contract, though. It may be useful to do so.

6. Speaking of the `PropertyObserver` interface, it was already a misnommer when returned from `expressionObserver`, it still is... `ValueObserver` perhaps?

7. I changed the `collectionObserver` contract so slightly by failing fast when receiving an invalid parameter, which is good practice for public APIs. I did so by moving the `throw` when `scope` is not an `Array` nor a `Map` outside of `subscribe`.

8. I do not know how to declare overloads properly with your typing transformer. `collectionObserver` should actually be typed as either `(Array|Map)` or `(Object, string)`.

9. Notifying the callback when a collection expression changes is done directly. I was not sure if I should enqueue a micro task or something. This ought to be fine since it's called from inside a change notification (so already inside a micro-task, right?).

10. I invented a new array change record type: `"reset"`, it's used when the array completely changed. It seems the array observation code does not quite do that when, say, an array is sorted but I couldn't figure it out. It's valuable to have such a notification, because on massive changes an array diff is just too costly and too big to be efficient or useful.

There is another API I would like to introduce: `dependencyObserver(scope, path)` with a specific use-case in mind, but I will stop here for now. (This API would be useful for scenarios such as `@computedFrom` or aurelia-validation `computedFrom`).

### DIFF
--- a/src/binding-engine.js
+++ b/src/binding-engine.js
@@ -45,39 +45,72 @@ export class BindingEngine {
       lookupFunctions);
   }
 
-  propertyObserver(obj: Object, propertyName: string): PropertyObserver {
-    return {
-      subscribe: callback => {
-        let observer = this.observerLocator.getObserver(obj, propertyName);
-        observer.subscribe(callback);
-        return {
-          dispose: () => observer.unsubscribe(callback)
-        };
-      }
-    };
-  }
-
-  collectionObserver(collection: Array|Map): CollectionObserver {
-    return {
-      subscribe: callback => {
-        let observer;
-        if (collection instanceof Array) {
-          observer = this.observerLocator.getArrayObserver(collection);
-        } else if (collection instanceof Map) {
-          observer = this.observerLocator.getMapObserver(collection);
-        } else {
-          throw new Error('collection must be an instance of Array or Map.');
-        }
-        observer.subscribe(callback);
-        return {
-          dispose: () => observer.unsubscribe(callback)
-        };
-      }
-    };
-  }
-
-  expressionObserver(scope: any, expression: string): PropertyObserver {
+  observer(scope: Object, expression: string): PropertyObserver {
+    // Fast path if the expression is just a property on scope.
+    // The Regex is not perfect but shortcuts all usual cases, notably it lacks unicode letters.
+    if (/^[$A-Za-z_][$A-Za-z0-9_]*$/.test(expression)) {
+      return {
+        subscribe: callback => {
+          let observer = this.observerLocator.getObserver(scope, expression);
+          observer.subscribe(callback);
+          return {
+            dispose: () => observer.unsubscribe(callback)
+          };
+        },
+        evaluate: () => scope[expression]
+      };
+    }
+    // Slow path: observe any expression
     return new ExpressionObserver(scope, this.parser.parse(expression), this.observerLocator);
+  }
+
+  collectionObserver(scope: Array|Map|Object, expression?: string): CollectionObserver {
+    // If called with just scope, we assume it is a collection that we want to directly observe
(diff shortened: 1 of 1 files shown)


jods4:
BTW sorry, I don't know what the diff is totally different. My guess is that it has to do with line-endings or tabs :(

jdanyow:
@jods4- thanks for the PR.  We're in the middle of some heavy refactoring in templating/binding/etc- will look at this after we get that in.

EisenbergEffect:
@jods4 I'm going to close this PR for now. We've made some serious refactorings in there. Take a look and if you want consider submitting a new PR on top of the new stuff.

jods4:
@EisenbergEffect So from what I see the API surface exposed is basically the same as before, but wrapped in a `BindingEngine` class.

My PR does not match the new code but is rather trivial to port to the new code.

The point it tackled where mostly public API design and I think my comments in #194 still apply.

This PR did the following things, which can still easily be done to `BindingEngine`:
- `propertyObserver()` and `expressionObserver()` can easily be merged in a single entry point `observer()` that accepts any expression, with a fast path if the expression is just a property. It will be easier to use and understand.
- Similarly, `collectionObserver()` could be extend to accept any path to an expression. Which will have a nice similarity with the `observer()` API above and is quite useful.

EisenbergEffect:
@jdanyow Your feedback is needed here. It would be a breaking change...so only if it can be completed by EOB Friday. Decision is up to you.

jdanyow:
Apologies, I don't have time to work through all this, add tests, etc, before tomorrow with everything else that needs to get done for the beta.  We could make an additive change to the BindingEngine API in the future.  

We should consider whether it would be useful for this omnipotent observer API to take into account the things we've added to the repeat in terms of expression unwrapping to enable observing collection mutation through value-converter boundaries.  

It might also be beneficial to explore this API in a plugin and gather feedback from the rest of the community like we're doing with aurelia-async.

Does this sound reasonable?

jods4:
I agree with @jdanyow that this API seems a bit rushed right now. There are many use-cases and many details that are not trivial to decide upon. Like observing paths, expressions, observing collections, observing dependencies only, handling behaviours (like signal), handling value converters...

I think it's a bit early to set this in stone and it needs more experimentations. Iterating on a plugin with feedback from more people seems like a very good idea.

EisenbergEffect:
Absolutely. We can always add things in the future. With the Beta, the main thing we want is to ensure no more breaking changes. We can still add new features for Beta 2. By the time we hit Beta 2, I want to have a feature freeze so that we can work solely on bug fixes, performance, quality for the v1 release. After that, we can add more new features again for v1.1 (but no breaking changes).

PR #226, opened 2015-11-13
Deleting last array element generates incorrect change record.

kevmeister68:
Hello Guys, this is my first Pull Request so I hope you'll be lenient on me :-)

I found a bug whereby deleting the last item of an array results in a repeat.for view actually removing the second-last item not the last item. 

Kevin

### DIFF
--- a/src/array-observation.js
+++ b/src/array-observation.js
@@ -65,7 +65,6 @@ class ModifyArrayObserver extends ModifyCollectionObserver {
     };
 
     array['splice'] = function() {
-      var methodCallResult = arrayProto['splice'].apply(array, arguments);
       var index = arguments[0];
       if (index >= array.length && array.length > 0) {
         index = array.length - 1;
@@ -74,6 +73,7 @@ class ModifyArrayObserver extends ModifyCollectionObserver {
       } else if (index < 0 ) {
         index = array.length + index - 1;
       }
+      var methodCallResult = arrayProto['splice'].apply(array, arguments);
       observer.addChangeRecord({
        type: 'splice',
        object: array,


EisenbergEffect:
@jdanyow Can you review? @kevmeister68 Would it be possible for you to add a unit test that produces the bug in order to show this fix resolves it?

martingust:
@EisenbergEffect @kevmeister68 

This will fix removing last item in array but will break for negative splices like: `this.items.splice(-1, 0, 'Foo');`. 

Changing `if (index >= array.length && array.length > 0)` to `if (index >= array.length)` should solve both.

But there is also negative splice without adding item `this.items.splice(-1, 0);` which will break in both.

This can most likely be handled in a more elegant way than I first implemented and that covers all scenarios.

EisenbergEffect:
@martingust Do you have time to get a full fix in today?

martingust:
@EisenbergEffect Unfortunately not tonight. I can probably sneak in 1-2 hours tomorrow to look at this.

kevmeister68:
I would add a unit test except I have no idea where they go, or how they work in terms of being testable against a visual outcome, unless the DOM is being explicitly checked afterwards. This was tested on our development codebase and found to fix the "deleting last item issue".

If there are problems with negative splices, it's not the problem I intended to fix. I think generally, however, it does not make a lot of sense to use a *modified* array as the basis for limiting your parameters in terms of constructing a change record, hence the patch.

Personally I think the changeRecord "handler" should be making the decisions on how to handle the splice outcome more so than this piece of code, otherwise it is still at the mercy of receiving a bogus changeRecord, if it happened to be injected from elsewhere.

martingust:
@kevmeister68 Agree, should probably not be in the array patch. I moved this to the `ModifyCollectionObserver` made some modification. The PR is here https://github.com/aurelia/binding/pull/229

If you wanted to take a look or try it out it would be great. And thanks for your PR and feedback.

EisenbergEffect:
Fixed in this PR: https://github.com/aurelia/binding/pull/229

PR #229, opened 2015-11-14
fix(ModifyCollectionObserver): handle out of bound splices

martingust:
@EisenbergEffect @jdanyow Please see if this looks alright to you guys. I think I got all the scenarios I could come up with right this time and move the logic to the `ModifyCollectionObserver` instead. There might be more scenarios though. But this will also fix regression bug when deleting last item or array. If any issue I can take a look again in about 4-5 hours.

### DIFF
--- a/src/array-observation.js
+++ b/src/array-observation.js
@@ -66,18 +66,10 @@ class ModifyArrayObserver extends ModifyCollectionObserver {
 
     array['splice'] = function() {
       var methodCallResult = arrayProto['splice'].apply(array, arguments);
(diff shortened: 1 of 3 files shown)


EisenbergEffect:
Great work. You and Jeremy know this part of the codebase better than I do now. So, since you both say it's good and we've got tests, I'll merge it in today. Thanks!

EisenbergEffect:
Clsoing this. The PR was merge. Github just didn't recognize it for some reason.

PR #253, opened 2015-12-05
perf(binding): reduce target property overhead

jdanyow:
**Memory**
memory

**CPU**
cpu

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -18,6 +18,7 @@ gulp.task('build-index', function(){
     'camel-case.js',
     'scope.js',
     'connectable-binding.js',
+    'connect-queue.js',
     'subscriber-collection.js',
     'array-change-records.js',
     'map-change-records.js',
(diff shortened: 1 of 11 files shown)


jdanyow:
with deferred-connect:
deferred

EisenbergEffect:
Looks pretty significant. Why is the top total larger though?

jdanyow:
The time labeled `(idle)` will be higher if I took longer to click stop on the cpu trace.  Safe to ignore that line.  

It doesn't seem to be possible to compare the actual times *across the three runs*... due to environmental factors perhaps?  Take `cloneNode` or `insertBefore` for example, they report a different "self" time in each of the 3 runs, even though they're doing the same thing each time (same template, same number of children, etc).  I was looking at the change in cost for each function in relation to other functions *in the same test run*- eg `createObserversLookup` falls to the bottom and `cloneNode` climbs to the top.  

Also implied in the "deferred connect" change is the memory increase related to setting up observers is deferred until after the rendering is done and is a more gradual climb due to the throttling in our `ConnectQueue`.

EisenbergEffect:
Understood. So, is this all ready now, assuming we merge all PRs across the repos?

adriatic:
You are almost dealing with the effects of Heisenberg's uncertainty principles here - given the size of scheduler's time-slice

jdanyow:
@EisenbergEffect this is ready

PR #267, opened 2015-12-21
fix(subscriberCollection): handle cascading calls

### LINKED ISSUE #252: Observing getter property results in TypeError: context is not a function
I have a custom attribute class that has a @bindable property that I would like to watch for changes. I use the BindingEngine to create a PropertyObserver and use that to subscribe to the property in question. When the property changes, I get the following error:

```
Uncaught TypeError: context is not a function 
callSubscribers @ aurelia-binding.js:244
call @ aurelia-binding.js:3221
flushMicroTaskQueue @ aurelia-task-queue.js:110
(anonymous function) @ aurelia-task-queue.js:60
```

Here are some code snippets... some code removed for conciseness:

```
@customAttribute('some-widget')
export class SomeWidget {
  @bindable readOnly = false;

  attached() {
    this.bindingEngine.propertyObserver(this, 'readOnly').subscribe((readOnly) => {
      ...
    });
  }
}
```

```
  <textarea some-widget="read-only.one-way: isArchived"></textarea>
```

The exact same code works without issue if remove the @bindable property and just create the property on 'this'. Obviously, I don't get the value of 'readOnly' that I am binding to in the parent in that case, however.

jdanyow:
Fixes #252, Fixes jdanyow/aurelia-computed#12

Handles scenarios where observers have subscriptions that overflow into the 'rest' buckets and calling one observer's subscribers causes another observer's subscribers to be called

### DIFF
--- a/src/subscriber-collection.js
+++ b/src/subscriber-collection.js
@@ -53,8 +53,9 @@ function removeSubscriber(context, callable) {
   return true;
 }
 
-let tempContextsRest = [null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null];
-let tempCallablesRest = [null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null, null];
+let arrayPool1 = [];
+let arrayPool2 = [];
+let poolUtilization = [];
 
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
If tests are passing...please merge at will.

PR #273, opened 2015-12-28
feat(Array): observe array prototype methods

jdanyow:
@EisenbergEffect I'm working on tests now. Existing tests in binding and templating-resources are passing. Let me know what you think.

### DIFF
--- a/src/array-observation.js
+++ b/src/array-observation.js
@@ -2,94 +2,123 @@ import {projectArraySplices} from './array-change-records';
 import {ModifyCollectionObserver, CollectionLengthObserver} from './collection-observation';
 import {subscriberCollection} from './subscriber-collection';
 
-var arrayProto = Array.prototype;
+let pop = Array.prototype.pop;
+let push = Array.prototype.push;
+let reverse = Array.prototype.reverse;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Was there any particular reason to store the original prototype methods on the Array prototype? As a results you have to have an extra deference via `this`. Could we just store them in module local variables instead?

jdanyow:
hmm I don't know what I was thinking there. I've updated the pull request to capture the original prototype methods using local variables and added some tests to ensure the array mutation methods function properly whether the array is observed or not. Let me know if it looks good to merge.

EisenbergEffect:
For perf reasons should the checks be more explicit?

```
if (this.__arrayObserver !== undefined) { ... }
```

jdanyow:
Looks like we can squeeze a little bit of perf out:  http://jsperf.com/falsey-check-vs-explicit-check

Will update the PR

jdanyow:
PR updated :zap:

EisenbergEffect:
Looks good. How do you feel about it? If you're good, then please go ahead and do the merge :)

PR #276, opened 2016-01-04
Support expression observation in the computedFrom

nyxtom:
In the `computedFrom` decorator, currently it is not possible to use the following syntax:

```
@computedFrom("myItem.itemsChanged")
```

Apart from this, I have a use case where it would be nice to have a referenced item in the template:

```
<template>

    <my-custom-list ref="customList"></my-custom-list> 

</template>
```

and in the code base:

```
@computedFrom("customList.itemsChanged")
get routeForCreate() {
    ///...etc
}
```

Doing the above would help greatly reduce the need to both observe when "customList" is available and consequently observe the "itemsChanged" on that custom list.

### DIFF
--- a/src/computed-observation.js
+++ b/src/computed-observation.js
@@ -1,4 +1,5 @@
 import {subscriberCollection} from './subscriber-collection';
+import {ExpressionObserver} from './binding-engine';
 
 const computedContext = 'ComputedPropertyObserver';
 
@@ -9,6 +10,7 @@ export class ComputedPropertyObserver {
     this.propertyName = propertyName;
     this.descriptor = descriptor;
     this.observerLocator = observerLocator;
+    this.parser = this.observerLocator.parser;
   }
 
   getValue(){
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@jdanyow Please review and see how this would work with similar issues as well as our computed plugin. Let's come up with a plan to address these issues officially.

EisenbergEffect:
@jdanyow Can you review this and if it's not quite right would it make sense to come up with a plan to add support for this?

jdanyow:
@nyxtom nice work, thanks for taking a look at this. There's a couple changes we should look into for the sake of performance.

1. Should we parse the expression when the computedFrom decorator is invoked instead of each time the property is bound?
2. Would it be more performant to use a standard observer when the expression is determined to be an instance of the AccessScope class?

Unfortunately the benchmark app has fallen into a bit of disrepair, let me get that fixed up and then we can continue on this PR.

EisenbergEffect:
@jdanyow What's our current thinking on this?

jdanyow:
I've started working on this, haven't finished, there's some refactoring to do because the ExpressionObserver used in this PR doesn't have the context/callable signature that the rest of the internal observers have because it was built for the BindingEngine.

jdanyow:
@nyxtom closing this- needed to go in a slightly different direction in addition to some refactoring of the binding primitives. In any case, thanks for the contribution.

PR #277, opened 2016-01-04
Set observation

martingust:
@EisenbergEffect @jdanyow If looks good this can be merged. This is required by https://github.com/aurelia/templating-resources/pull/178


jdanyow:
:rocket:

Issue #290, opened 2016-01-19
CSS (style) binding replaces whole style attribute instead of just modifying it


michalru3ch:
I have a scenario where jQuery sets some styles on an element (position, size etc.) and I want to use Aurelia binding to additionally set a color. But what happens is that Aurelia binding removes all style settings and I end up with just the color.

Instead of doing this `this.element.style.cssText = newValue;` StyleObserver should be implemented as KO does it (https://github.com/knockout/knockout/blob/master/src/binding/defaultBindings/style.js) - property by property, not as a whole style value.

Would you consider changing the implementation? Thank you in advance.

PR #293, opened 2016-01-23
Change only style properties set through observer/binding

z-brooks:
https://github.com/aurelia/binding/issues/290

Existing tests still pass. Tested in a skeleton as well.

### DIFF
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -51,31 +51,47 @@ export class StyleObserver {
   constructor(element, propertyName) {
     this.element = element;
     this.propertyName = propertyName;
+    
+    this.styles = [];
   }
-
+  
   getValue() {
     return this.element.style.cssText;
   }
-
+  
   setValue(newValue) {
-    if (newValue instanceof Object) {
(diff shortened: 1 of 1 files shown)


jdanyow:
@z-brooks :muscle: thanks! :sparkles:

stoffeastrom:
@jdanyow Is it possible to change a property in a style object and it will update? Doesn't seem to work, should it? It's a bit tedious to update the whole style object if it's only one property that needs to change..

jdanyow:
Agreed, that will require some more work. I wonder if this would be better:
```html
<div class.my-class.bind="foo"
     style.background-color.bind="backgroundColor">
</div>
```
Than:
```html
<div class.bind="{ 'my-class': foo }"
     style.bind="{ 'background-color': backgroundColor }">
</div>
```

thoughts?

stoffeastrom:
Yes that seems really nice!

EisenbergEffect:
I could go for that. Can we plug that into our parser in a simple way without affecting other things negatively?

jdanyow:
nice! will look into it

Issue #299, opened 2016-02-01
Variables are not being coerced into booleans on checked.bind


djedi:
If I have a value of `1` for example, bound to an input checkbox, checked.bind is not checking the checkbox. 

See this plunker: http://plnkr.co/edit/clvEGWYlA55E33WCkSVC?p=preview

EisenbergEffect:
Closing since this was fixed.

PR #300, opened 2016-02-01
Coerce value to boolean

djedi:
Fix for https://github.com/aurelia/binding/issues/299

### DIFF
--- a/src/checked-observer.js
+++ b/src/checked-observer.js
@@ -55,7 +55,7 @@ export class CheckedObserver {
 
     element.checked =
       isRadio && !!matcher(value, elementValue)
-      || !isRadio && value === true
+      || !isRadio && Boolean(value) === true
       || !isRadio && Array.isArray(value) && !!value.find(item => !!matcher(item, elementValue));
   }
 


z-brooks:
I think `Boolean([...])` is always *true*. Might be better to move that cast after the array check?

jdanyow:
@djedi awesome!

We'd love to merge this PR but will first need you to do a couple things:

1. amend the commit message (our release notes are automated based on a simple commit message format) 
2. sign the CLA

Instructions can be found in the `contributing.md` file in any of the repos.

Much appreciated!

jdanyow:
@z-brooks good point

@djedi please run the tests: `karma start` (instructions in readme.md)

fyi: 

result

Also, not sure if you're aware, but the value will be coerced, meaning 0 will end up becoming false on your property.  You may just want to use a value converter.

EisenbergEffect:
@djedi We will need a CLA sig before we can move ahead with this.
@jdanyow Does this still need fixes?

jdanyow:
@EisenbergEffect yes it does

EisenbergEffect:
@jdanyow Where are we at on this? Recommendations?

jdanyow:
Going to close this.  It needs work to address the issue @z-brooks mentioned, tests, etc.  I'm on the fence with the hard-coded coercion, it's caused a bunch of problems/confusion in input.value binding, need to take a step back and come up with a solution that will work across the binding system.

z-brooks:
gave this a shot before the close. It's fine if none of this is usable

https://github.com/z-brooks/binding/blob/master/src/checked-observer.js#L59
Does preserving the number value matter? Too ugly?
https://github.com/z-brooks/binding/blob/master/src/checked-observer.js#L83
Test.
https://github.com/z-brooks/binding/blob/master/test/checked-observer.spec.js#L185

PR #319, opened 2016-02-20
fix(*-observation): allow binding extended Map/Set

gingerik:
As far as I understood it is not possible to extend Maps & Sets using the Babel transpiler (at least for now). That's  too bad, as I would like to create a class similar to this:
```
export NumericSet extends Set {
  constructor() {
  }

  add(val) {
    if (typeof val === 'number') {
      super.add(val);
    }
  }
}
```

As expected, when I transpile this using Babel I get the following error:
```
TypeError: Method Set.prototype.add called on incompatible receiver #<NumericSet>
```

I tried to overcome this issue by overwriting the `add` method of my Set instance directly. Unfortunately, this does not work either, as Aurelia overwrites my `add` method using the native `Set.prototype.add`.

This PR could help me out be calling the `add` method directly rather than using the prototype by default. I've changed the `ModifyMapObserver` in a similar fashion.

### DIFF
--- a/src/map-observation.js
+++ b/src/map-observation.js
@@ -16,28 +16,40 @@ class ModifyMapObserver extends ModifyCollectionObserver {
   static create(taskQueue, map) {
     let observer = new ModifyMapObserver(taskQueue, map);
 
+    let methods = ['set', 'delete', 'clear'];
+    if (methods.find(m => mapProto[m] !== map[m])) {
+      mapProto = {};
+      methods.forEach(m => mapProto[m] = map[m]);
+    }
+
     map['set'] = function () {
+      let hasValue = map.has(arguments[0]);
+      let type = hasValue ? 'update' : 'add';
       let oldValue = map.get(arguments[0]);
-      let type = typeof oldValue !== 'undefined' ? 'update' : 'add';
       let methodCallResult = mapProto['set'].apply(map, arguments);
-      observer.addChangeRecord({
-        type: type,
-        object: map,
-        key: arguments[0],
-        oldValue: oldValue
-      });
+      if (!hasValue || oldValue !== map.get(arguments[0])) {
+        observer.addChangeRecord({
+          type: type,
+          object: map,
+          key: arguments[0],
+          oldValue: oldValue
+        });
+      }
       return methodCallResult;
     }
 
     map['delete'] = function () {
+      let hasValue = map.has(arguments[0]);
       let oldValue = map.get(arguments[0]);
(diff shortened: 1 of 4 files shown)


EisenbergEffect:
It's not possible to extend built-ins at all without runtime support from the browser itself.

gingerik:
I'm aware of that @EisenbergEffect. Maybe I didn't make my issue clear enough.
Even if I _could_ extend Sets, the `ModifySetObserver` would still overrule my custom `add` method.
For now, I'm trying to use a factory like this:
```
function createNumericSet() {
  let s = new Set();
  s.add = function(v) {
    if (typeof v === 'number') {
      Set.prototype.add.call(this, v);
    }
    return this;
  }
  return s;
}
```
But when I bind to such a Set, my own `add` method is being ignored due to a direct call to `Set.prototype.add`.

EisenbergEffect:
@jdanyow Any recommendation on how we can solve this?

jdanyow:
Looks like the PR would resolve @gingerik's issue at the expense of an extra object allocation for every ModifySetObserver is instantiation.  We can probably avoid taking that hit for most cases if we do something [like this](https://gist.run/?id=84be82e5430c6036a602)

thoughts?

EisenbergEffect:
First off...mega points on using your new tool to show the sample :)
Second, yes, that sounds like a reasonable approach.

jdanyow:
@EisenbergEffect lol- thanks!

@gingerik if that sounds good to you would you mind updating the PR with the perf changes, adding the Map tests (if they don't already exist) and signing the CLA if you haven't- pretty sure you've submitted a PR before but mentioning just in case.

gingerik:
That makes a lot of sense @jdanyow - I've just updated the PR by:
* adding the performance changes for sets & maps;
* setting the change record value to the value as added to the Set/Map (which can be different from the one in the argument);
* checking whether keys exists in a Map (I'm not sure this is desired, but I ran into this when copying the tests from `set-observation.spec.js` - With this PR there is a difference between having an `undefined` value or not having a value at all);
* adding map tests similar to the set tests.

And you're correct, I've signed the CLA before.

jdanyow:
@gingerik wow - that was quick, thanks!  Will merge in the next couple of days, I need to build and run it through some tests in templating-resources and the benchmarking app and then it will be ready to ship :ship:

gingerik:
You're welcome @jdanyow  - I'm happy to contribute to such a great project!

EisenbergEffect:
@jdanyow Ready to merge or still need to investigate?

PR #323, opened 2016-02-23
fix (observer-locator): observer locator for collections must return same instance of Observer for same instance collection

DrUNE:
I have noted that for Array instance Observer created in two distinct private properties. Even more, those instances where not the same depending on which API you use to create and retrieve them. This PR fixes management of observers for Collection types by ensuring that for one given instance of collection always will be only one MutationObserver instance.


EisenbergEffect:
@jdanyow Please view :)

jdanyow:
@DrUNE great PR :muscle:  much appreciated:exclamation:

DrUNE:
Glad it was useful :)

Issue #270, opened 2015-12-21
Binding should expose an `observable` decorator


PWKad:
Should be a subset of the `bindable` functionality that can be used on any class instance field -

```language-javascript
import {observable} from 'aurelia-binding';

export class MyViewModel {
  user;
  constructor() {
    this.user = new User();
  }
}

class User {
  @observable name = '';
  nameChanged(newValue) {
    console.log(newValue);
  }
}
```

jdanyow:
I like this idea! 

[...]

PR #327, opened 2016-02-28
feat(decorator-observable): add observable property decorator

Scapal:
Implement #270

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -46,7 +46,8 @@ gulp.task('build-index', function(){
     'listener-expression.js',
     'name-expression.js',
     'binding-engine.js',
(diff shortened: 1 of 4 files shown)


jdanyow:
:muscle: thanks @Scapal!  Can you confirm this works in the typescript skeleton? I did a quick test in the [typescript playground](http://www.typescriptlang.org/Playground) and it didn't look like it was working in TS.

@EisenbergEffect CLA check

Scapal:
@jdanyow now it works in the typescript skeleton (and still works using Babel)

jdanyow:
@Scapal thanks!

glen-84:
Is there anything similar for observing changes to array elements?

PR #328, opened 2016-03-01
feat(ref): support expressions

### LINKED ISSUE #214: .ref only supports properties on context
It would be nice to be able to do things like `<datatable datatable.ref='section2.datatable'>`.

But this currently doesn't work as `NameExpression` only supports properties directly on context (it does `context[property] = value`).

As assigning values to path is already supported by Aurelia (e.g. in bindings), I guess it shouldn't be too hard to reuse the same setter code for `NameExpression`. 

It would be the attributes behaviour more consistent and more powerful.

jdanyow:
fixes #214

### DIFF
--- a/src/name-expression.js
+++ b/src/name-expression.js
@@ -9,14 +9,14 @@ function getAU(element) {
 }
 
 export class NameExpression {
-  constructor(property, apiName) {
-    this.property = property;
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
Looks pretty good. Should we null out the source in the unbind?

PR #329, opened 2016-03-01
feat(BindingEngine): option controlling AccessMember assign behavior

### LINKED ISSUE #205: Consider adding an option to control AccessMember's automatic object creation
`AccessMember` [automatically creates an object](https://github.com/aurelia/binding/blob/0.10.2/src/ast.js#L189) when assigning a value to a property on an object that doesn't exist.

[Some people love this behavior](https://github.com/aurelia/binding/issues/129).  [Other people don't like it](https://github.com/aurelia/binding/issues/204).

Maybe we can make the behavior configurable?

cc @EisenbergEffect, @danfma

jdanyow:
fixes #205

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -1,4 +1,5 @@
 import {getContextFor} from './scope';
+import {options} from './options';
 
 export class Expression {
   constructor(){
@@ -249,7 +250,7 @@ export class AccessScope extends Expression {
 }
 
 export class AccessMember extends Expression {
-  constructor(object, name){
+  constructor(object, name) {
     super();
 
     this.object = object;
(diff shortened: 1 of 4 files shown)


jdanyow:
How does this look (aside from needing tests)? Can anyone think of a better name for the option?

Public API is the options property on the binding engine instance.

EisenbergEffect:
@jdanyow Should we log something if the settings is "false" but the instance doesn't exist?

jdanyow:
Tough call- I think some people (myself included) would think we should log when the object is automatically created as opposed to when assignment is skipped. There are developers on both sides of this issue.  Here's an example of the most common way this issue crops up: https://gist.run/?id=73ef440e9efde8b164ca

EisenbergEffect:
I suppose we could also have a logging setting. What do you think?

EisenbergEffect:
@jdanyow Where are we at on this also considering the binding behavior idea.

jdanyow:
closing- this is not up to date with the current plan outlined in https://github.com/aurelia/binding/issues/205

PR #342, opened 2016-03-09
Log warning when property can't be defined

gingerik:
First of all, this PR is not complete.

For some types of objects I wish not to allow adding properties after initialization (using `Object.preventExtensions()`). However, this also prevents Aurelia from observing these properties, as defining the required extra properties using `Object.defineProperty` fails. Currently, this is being ignored silently. This purpose of this PR is to add logging to the situations where this might occur.
I've you find this PR useful, I can add loggers for the other `Object.defineProperty` occurrences as well.

By the way, I actually find `Reflect.defineProperty` to be a more elegant solution than try & catching. But as this requires an extra polyfill in aurelia-polyfills, I hereby first mention it as a suggestion.

### DIFF
--- a/config.js
+++ b/config.js
@@ -15,6 +15,7 @@ System.config({
 
   map: {
     "aurelia-dependency-injection": "npm:aurelia-dependency-injection@1.0.0-beta.1.1.2",
+    "aurelia-logging": "npm:aurelia-logging@1.0.0-beta.1.1.2",
     "aurelia-metadata": "npm:aurelia-metadata@1.0.0-beta.1.1.3",
     "aurelia-pal": "npm:aurelia-pal@1.0.0-beta.1.1.1",
     "aurelia-pal-browser": "npm:aurelia-pal-browser@1.0.0-beta.1.1.2",
--- a/package.json
+++ b/package.json
@@ -27,11 +27,13 @@
       "dist": "dist/amd"
     },
     "peerDependencies": {
+      "aurelia-logging": "^1.0.0-beta.1.1.2",
       "aurelia-metadata": "^1.0.0-beta.1.1.3",
       "aurelia-pal": "^1.0.0-beta.1.1.1",
       "aurelia-task-queue": "^1.0.0-beta.1.1.0"
     },
     "dependencies": {
+      "aurelia-logging": "^1.0.0-beta.1.1.2",
       "aurelia-metadata": "^1.0.0-beta.1.1.3",
       "aurelia-pal": "^1.0.0-beta.1.1.1",
       "aurelia-task-queue": "^1.0.0-beta.1.1.0"
--- a/src/array-observation.js
+++ b/src/array-observation.js
@@ -123,7 +123,7 @@ class ModifyArrayObserver extends ModifyCollectionObserver {
   static for(taskQueue, array) {
     if (!('__array_observer__' in array)) {
       let observer = ModifyArrayObserver.create(taskQueue, array);
-      Object.defineProperty(
+      Reflect.defineProperty(
         array,
         '__array_observer__',
         { value: observer, enumerable: false, configurable: false });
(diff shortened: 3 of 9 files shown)


jdanyow:
I think this is a good idea, on the other hand it might result in a lot of log messages so we may need a way to turn it off.  We may also want to look into whether we can get rid of the try-catch altogether. Could help with performance.

EisenbergEffect:
Sounds good. @jdanyow  Let's take this as a starting point and see if we can make some recommendations to @gingerik to get it ready to merge in.

gingerik:
Any update on this @jdanyow @EisenbergEffect? Currently, I'm updating a project to Babel 6 and this really would have helped me doing so.

EisenbergEffect:
@jdanyow Where are we at on this? We need to wrap up the remaining binding issues if we can in the next few weeks.

jdanyow:
Looks good to merge except the missing tests.  @gingerik any interest in updating the PR with some tests? If not, no worries, I can take it from here.

gingerik:
@jdanyow I am willing to rebase/update/finish this PR. But two questions remain:
* Should the logging be configurable or not?
* Can I replace the `Object.defineProperty` instances with `Reflect.defineProperty` (and create a PR for aurelia-polyfills)?

jdanyow:
>* Should the logging be configurable or not?

Let's skip it for the first pass assuming we don't think it will impact most projects.  If there's a valid reason to want to turn it off that crops up we can add a config point.  Sound reasonable?

>* Can I replace the `Object.defineProperty` instances with `Reflect.defineProperty` (and create a PR for aurelia-polyfills)?

This sounds like a nice change.  Can you put together a quick jsperf that compares the existing try-define-catch approach with the if-reflectdefine-else approach?  If it looks good, please proceed with the changes and then I'll run the before and after through the benchmark suite to double confirm.  Should get a perf boost by removing the try-catch.  If so we could make the same change in observer-locator.js in the `createObserversLookup` method.

gingerik:
@jdanyow jsperf.com seems to be down for a while now. I've created a [jsfiddle](https://jsfiddle.net/bm792m4e/6/) to show the difference in performance. Please verify you're seeing a big improvement as well and I will update the PR.

jdanyow:
@gingerik this is great!  I'm seeing `Reflect.defineProperty` outperform `Object.defineProperty` by a factor of 6 in Chrome and a factor of 31 in Edge!  This is going to be a significant performance win for the binding system.  Very nice work! :muscle:  Please proceed :rocket: 

Made some minor changes to enable running in edge without the console:  https://jsfiddle.net/bm792m4e/7/

jdanyow:
oh I see it's not as pronounced when if you comment out the colliding defines, but still a noticeable improvement nonetheless.

gingerik:
@jdanyow That's actually a good point. Didn't try that -- well at least the logging should perform better then ;-). I've already created a PR for aurelia-polyfills as we probably need that one to be updated as well. I'll continue by updating this PR.

EisenbergEffect:
@jdanyow Where are we at with this? Do we just need to merge the polyfills PR first and then this one?

jdanyow:
my bad, I dropped the ball on this one- @gingerik you're done with this right?  If so I'll get it merged

EisenbergEffect:
I merged the polyfill in so if this is good to go, feel free to merge it and let me know :)

gingerik:
It's ready to pull @jdanyow

jdanyow:
thanks @gingerik :+1:

doktordirk:
@jdanyow 
has anyone tried this with typescript. 

also `typeof target !== 'object``might not work as compared to 
```js
    // https://people.mozilla.org/~jorendorff/es6-draft.html#sec-object-type
    function IsObject(x) {
        return typeof x === "object" ? x !== null : typeof x === "function";
    }
```

gingerik:
Good point @doktordirk!

Issue #149, opened 2015-08-12
computedFrom doesn't support paths


jods4:
It is my understanding and my experience that `computedFrom` does not work with paths, like: `@computedFrom('user.firstName')`.

I think that this is a very hard limitation that should be lifted:
1. Modeling and using sub-properties is common once your application gets even moderately complex.
2. There is sometimes no good alternative to `computedFrom`.

My use case is a computed property that creates an array from complex computations and lookups from other arrays. Definitively not something you want to poll and it is too complex for the auto-observation plugin. On the other hand its dependency is trivial to express as it is a single, nested property.

gabor-farkas:
+1 :)
(btw, is there no other way to create model listeners (watchers in angular terms) in the controllers?)

gabor-farkas:
Well and another thing: if we write paths or expressions in @computedFrom, there's no error message, it simply doesn't do anyting. Neither does the documentation mention exactly what can you write in there.

davismj:
+1

StrahilKazlachev:
+1

andrevlins:
+1

waywardz:
++1

jods4:
+1 my own ticket along other people, just to attract attention here @EisenbergEffect 

We've used Aurelia for 6 months in a complex project and there is something crucial missing; this might be part of the solution.

In UI you commonly have dependencies/computations between various elements. Some simple examples: list elements are filtered by other inputs (commonly just a textbox), you display the total of a shopping cart, you display shipping options only when the user did not choose "pickup" and is not in a foreign country, and so on.

A modern application needs to be able to express such values declaratively, as a computation from their dependencies:
```ts
get filteredList() { return this.list.filter(x => x.name.startsWith(x.search)) ; }
get grandTotal() { return this.list.reduce((total, x) => total + x.lineTotal); }
get showShippingOptions() { return this.shipping !== 'pickup' && this.country === 'US'; }
```
I've used properties for all these examples, although there can be other representations.

**Aurelia needs a good story for those use-cases.**

It could be some good integration with Reactive programming patterns and/or libraries, which are suited for those kind of things (and more!). I think Rx is great for many advanced scenarios but maybe a little overboard for something as simple as the motivating examples I've shown here.

Coming from Knockout, the examples above would 'just work', thanks to the inherent capability of KO to automatically discover and watch the dependencies of any computations (method, property or otherwise).

[...]

PR #345, opened 2016-03-12
computedFrom annotation - support for dot property name notation

sormy:


jdanyow:
@sormy thanks for the PR- unfortunately we cannot accept this one because there's already work underway to address #149 using Aurelia's binding expression parser.  We really appreciate the contribution just the same.

Issue #147, opened 2015-08-11
parametrical computed / recursive observation (make methods observable)


davismj:
In knockout, I would bind to a function that computes a value and returns a string. The value I passed in would largely be static, like an index on an array, but the function would then look up observables to compute the final value and thus set up subscriptions on those values.

In Aurelia, there is nothing like this. I cannot make a function called `computeValueForIndex(idx)` that subscribes to (or even dirty checks!) other variables on the view model. I cannot use a a computed `get computeValueForIndex(idx)` because getters cannot accept parameters.

jdanyow:
Is this the scenario:

ko view:
```html
<span data-bind="text: computeValueForIndex(idx)"></span>
```
ko view-model:
```javascript
// when ko invokes this function it makes note of the two observables that are called and will update bindings using this function when either of the two observables change
computeValueForIndex(idx) {
  return this.someObservable()[idx] + this.otherObservable();
}
```

In other words, we need the equivalent of `@computedFrom` or an `aurelia-computed` for methods on the viewmodel (only properties are supported today).

davismj:
precisely

jdanyow:
todo: consider when working on #149 and #249

EisenbergEffect:
@jdanyow Should  we consider rolling in your computed plugin? and somehow sharing code with that to address this and other related issues?

davismj:
Why not into core?

jods4:
Something I've been musing is adding KO auto-detect as an optional observation mecanism into Aurelia.

I've been thinking about something could work like that (naming could be better):
```typescript
class ViewModel {
  // This turns x into a property with instrumented getter, 
  // so that read access are detected
  @dependency 
  x: number;

  // can be a property or a method, or any piece of code in fact.
  // autoDependencies means that during evaluation, @dependency reads
  // should be collected and turned into dependencies, 
  // as if they had been added with @computedFrom
  @autoDependencies
  computed() {
    return this.x;
  }
}
```

This is basically what KO does. It could be an interesting alternative to existing mechanisms (ideally provided as an optional module).

A nice API could allow: 
- detecting and returning the dependencies of any piece of code, such as 
```
let [result, dependencies] = AutoDependency.Evaluate(func)
```
- choose whether dependencies are re-computed at each evaluation or only the first one.

EisenbergEffect:
What was the status on this? Can we support `computedFrom` on methods?

jdanyow:
[...]

PR #346, opened 2016-03-13
feat(computedFrom): support expressions

### LINKED ISSUE #149: computedFrom doesn't support paths

jdanyow:
fixes #149

### DIFF
--- a/src/computed-observation.js
+++ b/src/computed-observation.js
@@ -1,62 +1,6 @@
-import {subscriberCollection} from './subscriber-collection';
-
-const computedContext = 'ComputedPropertyObserver';
-
-@subscriberCollection()
-export class ComputedPropertyObserver {
-  constructor(obj, propertyName, descriptor, observerLocator) {
-    this.obj = obj;
-    this.propertyName = propertyName;
-    this.descriptor = descriptor;
-    this.observerLocator = observerLocator;
-  }
-
-  getValue(){
-    return this.obj[this.propertyName];
-  }
-
-  setValue(newValue){
-    this.obj[this.propertyName] = newValue;
-  }
-
-  call(context) {
-    let newValue = this.getValue();
-    if (this.oldValue === newValue)
-      return;
-    this.callSubscribers(newValue, this.oldValue);
-    this.oldValue = newValue;
-    return;
-  }
-
-  subscribe(context, callable) {
-    if (!this.hasSubscribers()) {
-      this.oldValue = this.getValue();
-
-      let dependencies = this.descriptor.get.dependencies;
-      this.observers = [];
-      for (let i = 0, ii = dependencies.length; i < ii; i++) {
-        let observer = this.observerLocator.getObserver(this.obj, dependencies[i]);
-        // todo:  consider throwing when a dependency's observer is an instance of DirtyCheckProperty.
-        this.observers.push(observer);
-        observer.subscribe(computedContext, this);
-      }
-    }
-
-    this.addSubscriber(context, callable);
-  }
-
-  unsubscribe(context, callable) {
-    if (this.removeSubscriber(context, callable) && !this.hasSubscribers()) {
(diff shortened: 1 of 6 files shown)


EisenbergEffect:
Very interesting stuff. With this in place, what do you think about adding the auto-dependency checking stuff that you had in that other plugin? Is it easy to add on top of this now that it's there? Or is that completely different?

jdanyow:
Definitely a step towards integrating that. If we do add the aurelia-computed capability here it would probably need to be opt-in via a decorator right? The explicit nature of the decorator would allow us to throw errors when we are not able to parse the expression(s) in the getter function's toString result.

The other thing related to this I'm investigating is support for https://github.com/aurelia/binding/issues/147 by adding logic to `@computedFrom` that would rewrite this:
```js
@computedFrom('some', 'stuff')
myClassMethod(myArg) {
  ...
}
```
to the equivalent of something like this:
```js
_myClassMethod(myArg) {
  ...
}

@computedFrom('some', 'stuff')
get myClassMethod() {
  return _myClassMethod.bind(this);
}
```

Then it should "just work":tm: provided [this line](https://github.com/aurelia/binding/blob/1.0.0-beta.1.2.2/src/ast.js#L284) is added [here](https://github.com/aurelia/binding/blob/1.0.0-beta.1.2.2/src/ast.js#L390).

What do you think?

EisenbergEffect:
Sounds pretty good to me.

EisenbergEffect:
Should we rename `computedFrom` to `computed` so that if you use it without an args it will auto-compute the props? (We can keep computedFrom and just create an alias to computed and deprecate the old one.)

jdanyow:
`computed` sounds good to me.  Will create a PR for that and the `aurelia-computed` logic integration.

Update on https://github.com/aurelia/binding/issues/147: my plan didn't quite work.  The changes to the decorator:
```js
export function computedFrom(...rest) {
  return function(target, key, descriptor) {
+    if (!descriptor.get) {
+      let innerKey = '__' + key;
+      Object.defineProperty(target, innerKey, { enumerable: false, configurable: true, value: descriptor.value });
+      descriptor.get = function() { return this[innerKey]; };
+      descriptor.set = function(newValue) { this[innerKey] = newValue; };
+      delete descriptor.value;
+      delete descriptor.writable;
+    }
  
    descriptor.get.dependencies = rest;
    return descriptor;
  }
}
```

Basically say "when the dependencies change, notify the binding system that the decorated class method changed".  This doesn't work.  The class method doesn't ever change. The result of *invoking* the class method potentially has changed. [This check](https://github.com/aurelia/binding/blob/1.0.0-beta.1.2.2/src/binding-engine.js#L123) (which is a valid check) prevents the change event from bubbling up and causing the binding to reevaluate.

Will need to think about https://github.com/aurelia/binding/issues/147 a bit more while working on the rest of this stuff.

OK to merge what I have so far?

jdanyow:
this was merged but rebasing the branch prevented the PR from closing automatically.  Closing this manually...

sormy:
Does it work well? It looks like even previously working syntax without expressions stop to work well.

I got master, build it and use method like that:
```
  @computedFrom('offset', 'topOffset', 'hourHeight')
  get top() {
    console.log('event get top()')
    return this.topOffset + this.offset * this.hourHeight / 60;
  }
```

and the code fall back to dirty checking and eating all cpu.

PR #349, opened 2016-03-14
feat(Expression): enable cloning and rebasing

jdanyow:
Few changes in this PR:
1. Primary change: make each of the AST node's `accept` method return the result of the method call, enabling more interesting visitors. 
2. Move the Unparser to it's own file and add tests for it.
3. Add a visitor that can clone expressions: `ExpressionCloner`. Useful as a base class for visitors that do expression rewriting...
4. Add an expression rewriting visitor that "rebases" expressions: `ExpressionRebaser`. Useful for things like [this](http://stackoverflow.com/a/35893248/725866) or in view-factory concatenation

Not sure if this all belongs in aurelia-binding or if 3 and 4 might need to go in a plugin?


jdanyow:
usage example for http://stackoverflow.com/q/35887132/725866

```js
import {inject} from 'aurelia-dependency-injection';
import {Parser, rebaseExpression} from 'aurelia-binding';

function unbind(binding) {
  binding.sourceExpression = binding.baseExpression;
  binding.baseExpression = null;
};

@inject(Parser)
export class ExpressionBindingBehavior {
  constructor(parser) {
    this.parser = parser;
  }

  bind(binding, source, expression) {
    let sourceExpression = binding.sourceExpression;
    expression = rebaseExpression(this.parser.parse(expression), sourceExpression);
    binding.baseExpression = sourceExpression;
    binding.sourceExpression = expression;
    expression.unbind = unbind;
  }
}
```

```html
<input value.bind="model & expression:key">
```
(where `key` is a string containing a javascript expression like `foo.bar.baz`)

PR #365, opened 2016-03-30
fix(SelectValueObserver): select default option

### LINKED ISSUE #364: Binding on `<select>` makes a `selected disabled` option get skipped
The js doesn't seem to matter, just the view:

```
<select value.bind="mySelection">
  <option value="showthis" disabled selected>Show This</option>
  <option value="dummyoption">dummyoption</option>
  <option value.bind="item" repeat.for="item of anArray" >${item}</option>
</select>
```

In the above code, the `selected` item ("Show This") is skipped, and the `dummyoption` is selected instead, which is clearly not our intent.

If you remove the `value.bind="mySelection"` from the `<select>`, the correct behaviour occurs, but of course then we don't get the value available anywhere.

Using `aurelia-binding@1.0.0-beta.1.3.1`

z-brooks:
Select the option the `selected` attribute by default when the value is not defined.

Fixes #364

### DIFF
--- a/src/select-value-observer.js
+++ b/src/select-value-observer.js
@@ -64,7 +64,7 @@ export class SelectValueObserver {
     while(i--) {
       let option = options.item(i);
       if (clear) {
-        option.selected = false;
+        option.selected = option.defaultSelected;
         continue;
       }
       let optionValue = option.hasOwnProperty('model') ? option.model : option.value;
--- a/test/select-value-observer.spec.js
+++ b/test/select-value-observer.spec.js
@@ -38,6 +38,25 @@ describe('SelectValueObserver', () => {
     return value;
   }
 
+  describe('default selection', () => {
+    var obj, el, binding;
+    beforeAll(() => {
+        obj = { selectedItem: null };
+        el = createElement(
+            `<select>
+            <option value="A">A</option>
+            <option value="B">B</option>
+            <option selected value="C">C</option>
+            </select>`);
+        document.body.appendChild(el);
+        binding = getBinding(observerLocator, obj, 'selectedItem', el, 'value', bindingMode.twoWay).binding;
+    });
+      
+    it('selects default', () => {
+        expect(el.value).toBe('C');
+    });
+  });
+
   describe('single-select strings', () => {
     var obj, el, binding;
 


jdanyow:
Thanks @z-brooks - I'm conflicted about what to do with this issue.  When an element is data-bound, the model value is assigned to the view.  In #364, the model value can't be assigned to the view because the model value is invalid.  With your PR the view will look like what @egeland expects but the side effect will be the element value and the model value are still not in sync (so I guess we're no worse off in this case) OR the model value would end up getting updated to match the view (which I think would be bad or at least inconsistent with the way everything else works).  I haven't tested the PR to see what ends up happening.

I think the real fix for @egeland would be to start with a valid model value- something that matches one of the option values.  I think it's weird to data-bind something but at the same time expect the view to supply the initial value- nothing else in the binding system works like that.

@z-brooks / @egeland what are your thoughts?

egeland:
Thanks for the quick fix, @z-brooks (I also have not tested it yet)

@jdanyow , I can't imagine I'm the only one to want to put a 'header' option in and then supply the selectable options from a `repeat.for...`..
Letting us do it this way seems more intuitive to me (which is why I tried it). Of course, I would be happy with a solution that makes even more sense.

If that solution is to `unshift` another item with the 'header' info onto my options array, and set a `disabled.bind` based on a property, then so be it. But in any case, we should probably document the "official" solution somewhere.

As always, I really appreciate all the work you are all doing on Aurelia!

jdanyow:
Would something like this work for you?  https://gist.run/?id=581f7deebc59088bc9412427c7fc82d0

Essentially the model's initial value needs to match the value of the "header" option.

```js
export class App {
  selectedColor = '';
  colors = ['Red', 'Green', 'Blue']
}
```

```html
<template>
  <select value.bind="selectedColor">
    <option value="" disabled>Pick a color...</option>
    <option value="something">Dummy Option</option>
    <option repeat.for="color of colors" value.bind="color">${color}</option>
  </select>
  
  selectedColor = '${selectedColor}'
</template>
```

z-brooks:
Just an idea:
https://github.com/aurelia/binding/blob/master/src/select-value-observer.js#L88 is changed to
```es6
if (!option.selected || option.disabled ) {
    continue;
}
```
This should prevent any value in the disabled option from being set to the model. A disabled option probably shouldn't have a value?

The only issue might be if `disabled` is not the same as the `selected` option it will always switch back to the default. If that is a problem?
```html
<option disabled>one</option>
<option selected>two</option>
```
`one` can never be set, it will always change to `two`. If that is a problem?

egeland:
@jdanyow Yeah, I think that would work fine. So maybe if you could just add a note in the docs, maybe with this example shown, that would help others in the future, too.

EisenbergEffect:
Ok, looks like this is good to close. @jdanyow Can you add a note in the docs you are writing on binding? Thanks!

PR #366, opened 2016-03-30
chore(travis): add travis configuration

mbroadst:
This adds a simple TravisCI configuration to ensure that the project is always linted. If/when travis is enabled for this project, all commits/PRs will run travis jobs to catch early errors before code makes it into the main repo. Travis can also run karma tests via `xvfb` if we want to go down this road. 

/cc @bryanrsmith (I spoke with Rob and Jeremy about this, I believe you were working on CI at one point?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: node_js
+node_js:
+  - '0.12'
+  - '4'
+  - '5'
+
+sudo: false
+cache:
+  directories:
+    - node_modules
+
+script: gulp lint
+


mbroadst:
it should be noted that this repo currently has a _lot_ of linting errors :smile: so it should be a good place to start if this is indeed desired for all of the aurelia repos

Gheoan:
Linting is also a good way to reduce size. For example #367 removed more code then added (574 additions and 584 deletions). It is a bit but it counts :D.

Gheoan:
@mbroadst You might want to look at aurelia/router#258 and aurelia/router#260.

bryanrsmith:
Yes, I set up Travis in several repos a few months back. I'd love to get it set up on the others! I put together a set of instructions and a hacky script that should add all the files and necessary gulp changes here: https://gist.github.com/bryanrsmith/0e979409d494531eb2fe. I haven't reviewed the script since the recent build changes to use babel 6, so there's a chance it needs a little update.

PR #367, opened 2016-03-30
style(module): lint the entire module

mbroadst:


jdanyow:
:muscle: you are the man :clap:

mbroadst:
it probably pays to heed the lines including `// eslint-disable-line`, they tend to be things like assignments as part of logical expression etc. These are most likely harmless but could be confusing/dangerous

jdanyow:
All tests passing, also tested this in templating-resources test suite.

Thanks @mbroadst!

PR #403, opened 2016-05-14
Respect non-enumerable properties

gingerik:
When creating a property like this:
```
Reflect.defineProperty(obj, 'foo', {
  configurable: true,
  enumerable: false, // <-- non-enumerable
  set: setter,
  get: getter
});
```
the property observer will add its own setter and getter (which is a good thing), but also sets the `enumerable` setting to `true`.
This PR makes the observer respect the existing setting.

### DIFF
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -96,7 +96,8 @@ export class SetterObserver {
 
     if (!Reflect.defineProperty(this.obj, this.propertyName, {
       configurable: true,
-      enumerable: true,
+      enumerable: this.propertyName in this.obj ?
+          this.obj.propertyIsEnumerable(this.propertyName) : true,
       get: this.getValue.bind(this),
       set: this.setValue.bind(this)
     })) {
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
@jdanyow Please review.

src/property-observation.js
```diff
@@ -96,7 +96,7 @@ export class SetterObserver {
 
     if (!Reflect.defineProperty(this.obj, this.propertyName, {
       configurable: true,
-      enumerable: true,
+      enumerable: this.obj.propertyIsEnumerable(this.propertyName),
```
gingerik: It might be better to replace this with `this.propertyName in this.obj ? this.obj.propertyIsEnumerable(this.propertyName) : true`. But I'm not sure whether the property is always existent.

jdanyow: Right- there's no guarantee the property will be defined so the change you mentioned ^^^ would be necessary.  Could you add a test for that scenario as well?  Otherwise this looks great!

EisenbergEffect:
If we can get this in today, we can get the fix out tomorrow. Let me know.

gingerik:
I've updated the PR!

EisenbergEffect:
@jdanyow Can you do a final review?

jdanyow:
looks good, merged. thanks @gingerik!

PR #423, opened 2016-06-02
feat(ViewEngineHooksResource): Enable registering ViewEngineHooks by convention

davismj:
A ViewEngineHooks object can now be registered by either
 - Creating a {Name}ViewEngineHooks class that implements ViewEngineHooks
 - Decorating a class that implements ViewEngineHooks with @viewEngineHooks

and
 - Registering that class with the view via a <require> tag
 - Configuring the class as a globalResource


atsu85:
Did You include dist folder by accident?

davismj:
You're right I did.

PR #424, opened 2016-06-03
feat(view-engine-hooks-resource): Add default ViewEngineHooksResource and decorator

davismj:


EisenbergEffect:
@davismj This should actually go in the templating repo also. Do you mind updating your existing PR with this?

davismj:
i don't understand? there is a templating pr as well
https://github.com/aurelia/templating/pull/371

EisenbergEffect:
I mean the actual ViewEngineHooksResource type should go in the templating repo, not in the binding repo.

EisenbergEffect:
The binding engine is a lower dependency than templating. The actual view engine and all it's details are defined in templating, so technically binding shouldn't know anything about that. @jdanyow may have some thoughts on the camel case helper.

davismj:
Turns out the `camelCase` class was not required after all. Pull request updated here: https://github.com/aurelia/templating/pull/381

PR #425, opened 2016-06-05
Update package.json

MeirionHughes:
allow typescript to find npm module typings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "license": "MIT",
   "author": "Rob Eisenberg <rob@bluespire.com> (http://robeisenberg.com/)",
   "main": "dist/commonjs/aurelia-binding.js",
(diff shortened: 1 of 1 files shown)


MeirionHughes:
this needs to be done on pretty much all the projects btw.

EisenbergEffect:
I don't believe this will work in its present form. We need to change the format of the dts files as well. The team has been discussing this and making plans. @niieani can provide some commentary.

niieani:
Indeed, first we need to deal with this: https://github.com/aurelia/dependency-injection/pull/95.

MeirionHughes:
ahh yes, you've declared a module export; fair enough. Just as well I didn't go off and PR them all. :D

EisenbergEffect:
I'm going to close this for now, but know that we are working on this and planning to use the typings strategy. We hope to get this in place pretty soon.

Issue #269, opened 2015-12-21
bindingExpression updates the target unnecessarily for equal date values


sylvain-hamel:
`bindingExpression` compares old and new values using `===` and therefore will update the target if the source value changes from `new Date(1776, 7, 4)` to `new Date(1776, 7, 4)`.

See https://github.com/sylvain-hamel/binding/commit/d9e16e28adc413abe16756b388b15e36f4839457 for the unit tests that reproduce the issue.

After discussing with @jdanyow I'm opening this issue to keep track of this limitation.   Changing `bindingExpression` could be costly, using a custom binding behavior is a workaround if someone needs to avoid the unnecessary update to the target:

```
<my-element max-date="date1 & dateComparison", min-date="date2 & dateComparison">
```

jdanyow:
Discussed this with @sylvain-hamel last week in a private gitter chat. The binding engine's behavior is technically correct, notifying changes when a property's date instance changes, however this might is not optimal when the new date instance has the same date value as the original date instance. 

It could be argued that it's not the binding instance's responsibility to sort this out, we definitely don't want to add additional logic to this already hot path. Perhaps the component or logic that is assigning a new date instance to a property whose value is a date instance with the same date value should be updated to do a date comparison prior to assigning.

A workaround that would avoid adding overhead to non-date-related bindings would be a binding behavior.  Something like:

export class DateComparisonBindingBehavior {
  bind(binding, source) {
    binding.dateComparedUpdateSource = binding.updateSource;
    binding.updateSource = function(newValue) {
      let currentValue = this.sourceExpression.evaluate(this.source, this.lookupFunctions);
      if (!currentValue || !newValue || +currentValue !== +newValue) {
        this.dateComparedUpdateSource(newValue);
      }
    }
  }

  unbind(binding) {
    binding.updateSource = binding.dateComparedUpdateSource;
    binding.dateComparedUpdateSource = null;
  }
}

PR #428, opened 2016-06-08
fix(property-observation): now detects no change when both oldValue and newValue are NaN

nicolasdejong:
In Javascript the expression "NaN === NaN" always resolves to false.
Aurelia uses the expression "oldValue !== newValue" to detect if a property has changed and change-handlers should be called.
This fails if both oldValue and newValue are NaN.
Therefore an extra comparison was added: "!( isNaN(oldValue) && isNaN(newValue) )".

### DIFF
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -57,7 +57,7 @@ export class SetterObserver {
   setterValue(newValue) {
     let oldValue = this.currentValue;
 
-    if (oldValue !== newValue) {
+    if (oldValue !== newValue && !(Number.isNaN(oldValue) && Number.isNaN(newValue))) {
       if (!this.queued) {
         this.oldValue = oldValue;
         this.queued = true;
--- a/test/binding-engine.spec.js
+++ b/test/binding-engine.spec.js
@@ -51,6 +51,27 @@ describe('bindingEngine', () => {
     });
   });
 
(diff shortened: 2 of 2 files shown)


nicolasdejong:
This is not correct. isNaN('text') is true as well.

nicolasdejong:
typeof NaN === 'number'

Sometimes Javascript can be weird...

nicolasdejong:
Fixed by using Number.isNan

The tests now run correctly again. I forgot to remove the 'f' before 'it' in the text which made me miss this issue.

Gheoan:
Will the additional comparison affect performance?

nicolasdejong:
The performance penalty is negligible.

The old comparison: a !== b
Added: && !(Number.isNan(a) && Number.isNan(b))

Number.isNan(n) is implemented as: n !== n  (because NaN !== NaN)

So it becomes: a !== b && !( a !== a && b !== b)

The === or !=== operation is, as far as I know, very cheap.

And even if it wasn't, the change detection is incorrect without the added NaN check. I ran into this issue because of an endless loop of a two-way binding that got the value NaN.

jdanyow:
thanks for the PR- there's a whole bunch of places we'd need to add this logic.  I don't think the use case is compelling enough to warrant the changes.

Related to https://github.com/aurelia/binding/issues/269

PR #440, opened 2016-06-25
feat(ast): bindingGetter and bindingSetter

niieani:
Allow overriding the getter and setter used by the binding by declaring methods bindingGetter() and bindingSetter(value) as properties on the getter / setter respectfully.
This enables concepts such as [binding interception / proxying](https://gist.github.com/niieani/9300029424327b4aa282fa7f5a1e8b89) and thus in combination with Aurelia's Observers solves async bindings (Promises, Rx Observables / Subjects) and opens up new possibilities for binding and state-management extensibility.

The changes don't affect any previous code (no breaking changes).

Performance impact should be negligible (but should be tested on things making heavy use of bindings, like dbmonster), however if needed, there is still the possibility of caching the reference to the bindingGetter and bindingSetter methods per their respectful contexts.

Example use-cases (binding interception): https://gist.run/?id=e4185ddcdb80b67dbcbcd209689c6901

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -230,12 +230,12 @@ export class AccessScope extends Expression {
 
   evaluate(scope, lookupFunctions) {
     let context = getContextFor(this.name, scope, this.ancestor);
-    return context[this.name];
+    return getScoped(context, this.name);
   }
 
   assign(scope, value) {
     let context = getContextFor(this.name, scope, this.ancestor);
-    return context ? (context[this.name] = value) : undefined;
+    return context ? (setScoped(context, this.name, value)) : undefined;
   }
 
   accept(visitor) {
@@ -259,7 +259,7 @@ export class AccessMember extends Expression {
 
   evaluate(scope, lookupFunctions) {
     let instance = this.object.evaluate(scope, lookupFunctions);
-    return instance === null || instance === undefined ? instance : instance[this.name];
+    return instance === null || instance === undefined ? instance : getScoped(instance, this.name);
   }
 
   assign(scope, value) {
@@ -270,8 +270,7 @@ export class AccessMember extends Expression {
       this.object.assign(scope, instance);
     }
 
-    instance[this.name] = value;
-    return value;
+    return setScoped(instance, this.name, value); // eslint-disable-line
   }
 
   accept(visitor) {
@@ -686,11 +685,30 @@ function getFunction(obj, name, mustExist) {
   throw new Error(`${name} is not a function`);
 }
 
+function getScoped(obj, key) {
+  let descriptor = Object.getPropertyDescriptor(obj, key);
+  if (descriptor && descriptor.get && descriptor.get.bindingGetter) {
+    return descriptor.get.bindingGetter();
+  } else {
+    return obj[key];
+  }
+}
+
+function setScoped(obj, key, value) {
+  let descriptor = Object.getPropertyDescriptor(obj, key);
+  if (descriptor && descriptor.set && descriptor.set.bindingSetter) {
+    descriptor.set.bindingSetter(value);
+  } else {
+    return obj[key] = value;
+  }
+  return value;
+}
+
 function getKeyed(obj, key) {
   if (Array.isArray(obj)) {
     return obj[parseInt(key, 10)];
   } else if (obj) {
-    return obj[key];
+    return getScoped(obj, key);
   } else if (obj === null || obj === undefined) {
     return undefined;
   }
@@ -708,7 +726,7 @@ function setKeyed(obj, key, value) {
 
     obj[index] = value;
   } else {
-    obj[key] = value;
+    setScoped(obj, key, value);
   }
 
   return value;


atsu85:
@niieani, `@bindingIntercept` and `promiseBinding(...)` from the sample gist seem to be universialy usable - Do You have any plans to add them also to Aurelia (maybe using a plugin)?

niieani:
@atsu85 it all depends on the guys in charge ;) in this case @jdanyow. If this goes through, I don't see why wouldn't we include the `bindingIntercept` and its most common use-cases.

jdanyow:
@niieani let's see if we can come up with something less invasive.  I'm seeing a 4x (chrome) to 10x (firefox) reduction in speed in a simple benchmark:  https://gist.run/?id=574e579f0d91245da74e1d238e8b3fe6

I like the binding functions approach you and I have been discussing for several reasons:
* no special code required in view-models
* it uses capabilities we already have (especially attractive because we're in RC stage)
* it's an opt-in model- no overhead for apps or bindings that don't need it (arguably 99.9% don't need this)
* the binding expressions reflect what is actually happening. Someone looking at a view can easily understand what is going on.

https://gist.run/?id=e863f4affa5267538675911fbb8eac02

niieani:
The main question that I've been contemplating when designing this is: should the View be tightly coupled to the data-structure of the ViewModel, i.e. should the designer be aware of the fact that a value is a Promise or an Observable or anything else for that matter? 

In my opinion - no - and by defining a binding like: `async(firstName).value` - we are tightly coupling it to the underlying Promise implementation. If we decided to swap out the Promise for an Observable, or a static value, we'd have to go back to the View and make the change. By simply binding to: `firstName` - the implementation isn't hard coded in the View and could easily be swapped out.

Think about this scenario: In large projects with a clear separation of responsibilities, a designer could make a complete View template with a mock implementation of the ViewModel. The developer could then work only on the ViewModel without having to touch the View. I see that alone as a huge benefit. And, if in some time the website/app decides to make design changes, the designer doesn't need to run the full-blown app on his laptop, but can continue editing his View using his mock ViewModel implementation. The developer then swaps the View for the new file he got from the designer and in the best case scenario - they're ready to ship. The same works in the opposite direction - i.e. in case of a complete overhaul of the backend - the same View should work without changes.

Note that ViewConverters and BindingBehaviors don't introduce as much coupling as this would - in most cases they're used for things like converting Dates to easily-readable string (it's hard to imagine a Date object would be swapped out for a different construct), translation, or throttling-type operations -- all of which could be decisions delegated to the designer.

@EisenbergEffect I'd love to hear your thoughts on this -- not necessarily on the implementation details, as I understand you delegate that to @jdanyow, but on the "tight View-ViewModel coupling" vs "loose View-ViewModel coupling" - what aligns better with the Aurelia ecosystem?

I'm not going to insist on the specific implementation from this PR, but it would be good to somehow get the best of both worlds or reach a good compromise between speed, features and ease-of-development.

Going back to the implementation, I've indeed did my own set of benchmarks here: https://gist.run/?id=4cae4eb22b9cb34b8e6b2d55d8a6bb52

I've also tried an alternative implementation which I thought would be more performant, i.e. instead of installing the hook on the getter, installing it on a prefixed property, like this:
```js
  let proxyKey = bindingProxyPrefix + key;
  return proxyKey in obj ? obj[proxyKey] : obj[key];
```

But strangely enough, the speed is lower than when using `getOwnProperty`. I suppose concatenation of strings is that expensive. So, I've tried caching the concatenated strings -- improves the performance a lot, at least in Chrome.

```
script.js:214 Traditional get x 25,780,680 ops/sec ±6.18% (52 runs sampled)
script.js:214 BindingProxy property x 3,802,929 ops/sec ±3.25% (54 runs sampled)
script.js:214 BindingProxy property (cached string) x 10,304,400 ops/sec ±7.12% (55 runs sampled)
script.js:214 BindingGetter (own) x 3,106,321 ops/sec ±3.18% (52 runs sampled)
script.js:214 BindingGetter (all) x 3,040,578 ops/sec ±2.52% (57 runs sampled)
script.js:214 BindingGetter with cache (Map) (own) x 5,882,579 ops/sec ±6.79% (55 runs sampled)
script.js:214 BindingGetter with cache (Map) (all) x 6,340,303 ops/sec ±8.60% (47 runs sampled)
script.js:214 BindingGetter with cache (Object) (own) x 1,420,932 ops/sec ±4.50% (54 runs sampled)
script.js:214 BindingGetter with cache (Object) (all) x 1,391,135 ops/sec ±2.61% (52 runs sampled)
``` 

In Firefox I get the better performance with a Map cache of the getter, strangely enough caching the string doesn't help there:

```
Traditional get x 121,714,167 ops/sec Â±3.36% (31 runs sampled) script.js:214:3
BindingProxy property x 7,371,019 ops/sec Â±1.18% (28 runs sampled) script.js:214:3
BindingProxy property (cached string) x 6,043,873 ops/sec Â±1.28% (31 runs sampled) script.js:214:3
BindingGetter (own) x 1,413,548 ops/sec Â±4.90% (29 runs sampled) script.js:214:3
BindingGetter (all) x 1,356,953 ops/sec Â±6.70% (26 runs sampled) script.js:214:3
BindingGetter with cache (Map) (own) x 18,857,812 ops/sec Â±2.82% (26 runs sampled) script.js:214:3
BindingGetter with cache (Map) (all) x 19,196,718 ops/sec Â±1.74% (26 runs sampled) script.js:214:3
BindingGetter with cache (Object) (own) x 2,709,819 ops/sec Â±3.20% (24 runs sampled) script.js:214:3
BindingGetter with cache (Object) (all) x 2,734,071 ops/sec Â±7.21% (21 runs sampled) script.js:214:3
```

The real question is whether these benchmarks really mean anything in the big scope of things. The number of ops/sec is so large that we should make a more comprehensive test on whether this becomes a bottleneck or does it only have a minuscule effect on the performance.

Binding Functions as defined by the example you gave are okay for some use cases like promises, but they seem very hacky, plus they create tight-coupling between the View and the ViewModel which I mentioned above. They also require the original object to be modified (acceptable in the case of Promises, but unacceptable in some other cases, e.g. if the object is to be later stored in the database) and are prone to leaks - even with the dispose method in place -- and let me explain why: 

Since it's not possible to track which binding we are disposing of (`dispose()` in the unbindHook disposes of all of them), if you look at the code referenced, the `dispose` method will be called during unbinding of **any** of the Views that reference it. Meaning: all of the bindings will be disposed of, not just the one that created the hook. It's possible to do a ref-count and only dispose of them when it reaches 0, but that means leaks could still happen until all of the bindings are unbound, e.g. if there was an `if.bind` that periodically shows and hides a template that contains a reference to a binding that is also bound elsewhere, it would be leaking every time that template is unbound.

And then there's the two-way binding situation which still isn't possible with that sort of implementation. You can't feed new data **from** the binding **to a function** without it passing through to the ViewModel (i.e. without it invoking a setter). 

That's why I've arrived at the idea of "proxying" the data to-and-from the bindings to the ViewModel -- kind of like ValueConverters, but more flexible. 

Thoughts?

doktordirk:
re string concat: tried with computed strings? that.s supposed to be faster

niieani:
@jdanyow I've run DBMonster with the change from this PR - there is no perceivable difference in performance at all. I get the same FPS with and without it. I think benchmarks matter in the larger context, and if you look at the whole binding & rendering process the cost of this feature is negligible:

Chrome:
screen shot 2016-06-28 at 17 33 51

Firefox:
screen shot 2016-06-28 at 17 58 09

Profiling the DBMonster with aurelia-binding from this PR shows that the cost of `getScoped` is only **~1%** under Chrome and in Firefox only **0.48%** - and DBMonster is a binding-intensive setting. In real-life situations that's going to be a fraction of a percent. Given the potential and possibilities that this feature opens with regards to new ways of handling bindings, I think we should reconsider. Unless you have a better idea for how to implement two-way binding proxies? How about a way to make this opt-in-only?

@EisenbergEffect Are we going to handle async before RTM? 

As a side note: It would be awesome if we could have a performance CI running the latest DBMonster (and other perf. benchmarks) with every Aurelia release. We could then automatically draw charts to see how the performance is changing over time and visually track any potential regressions or improvements.

EisenbergEffect:
I'm not sure there's a way to make this happen in the next 3 weeks, so probably not before 1.0RTM. But I don't see why we couldn't turn around and do a 1.1 of the binding library once we feel good about the implementation, perf, etc. of an async solution. @jdanyow can comment more on this. We've got other benchmarks as well, not just db monster.

niieani:
@EisenbergEffect OK. Could you comment on the: **tight ViewModel-Model coupling** _vs_ **loose ViewModel-Model coupling** issue I put forth [a few comments back](https://github.com/aurelia/binding/pull/440#issuecomment-228604541) since it's more of a project-design issue than a technical one?

PR #441, opened 2016-06-29
Update binding-value-converters.md

rockResolve:
doc: binding-value-converters

fix small grammar error


EisenbergEffect:
Awesome. Thanks!

PR #475, opened 2016-08-07
typo in examples

ayurmedia:
should be Motherboard, CPU, Memory as Content


ayurmedia:
same as pullrequest from 	@nmyvision

Issue #406, opened 2016-05-17
Look into how #342 (unable to observe property logging) affects this scenario


jdanyow:
https://github.com/aurelia/binding/issues/342

``` js
import {children} from 'aurelia-framework';

export class TabControl {
  @children('tab-item') items = [];
  selectedItem = null;

  itemsChanged(mutations){
    if (this.items.length > 0) {
      if (!this.selectedItem || this.items.indexOf(this.selectedItem) === -1){
        this.selectItem(this.items[0]);
      }
    }
  }

  selectItem(item){
    if (this.selectedItem) {
      this.selectedItem.isSelected = false;
    }

    this.selectedItem = item;

    if (this.selectedItem) {
      this.selectedItem.isSelected = true;
    }
  }
}
```

``` html
<template>
  <ul class="nav nav-tabs">
    <li repeat.for="item of items" class="${item.isSelected ? 'active' : ''}">
      <a href="#" click.trigger="$parent.selectItem(item)">
        <span replaceable part="item-header">
          ${item.heading}
        </span>
      </a>
    </li>
  </ul>

  <content></content>
</template>
```

jdanyow:
```js
  // babel's decorator logic uses !!descriptor.configurable which creates read-only
  // properties that can't be observed with the SetterObserver.  Make sure the
  // property remains configurable unless it was explicitly set to false...
  if (descriptor.configurable !== false) {
    descriptor.configurable = true;
  }
```

PR #477, opened 2016-08-11
Fix issue 476

ccantill:
SetterObserver doesn't work if a property is already defined.

### DIFF
--- a/src/observer-locator.js
+++ b/src/observer-locator.js
@@ -144,14 +144,14 @@ export class ObserverLocator {
         if (existingGetterOrSetter.getObserver) {
           return existingGetterOrSetter.getObserver(obj);
         }
-
-        // attempt to use an adapter before resorting to dirty checking.
-        let adapterObserver = this.getAdapterObserver(obj, propertyName, descriptor);
-        if (adapterObserver) {
-          return adapterObserver;
-        }
-        return new DirtyCheckProperty(this.dirtyChecker, obj, propertyName);
       }
+
+      // attempt to use an adapter before resorting to dirty checking.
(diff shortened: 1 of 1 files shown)


jdanyow:
closing because the root cause is the way babel transpiles the decorated props- see https://github.com/aurelia/binding/issues/406 for more info.

doktordirk:
@jdanyow @EisenbergEffect
 a) i think the issue is more https://github.com/loganfsmyth/babel-plugin-transform-decorators-legacy/pull/44 & https://github.com/tc39/proposal-class-public-fields/issues/36
b) since you tell everyone to use that implementation of decorators, you can.t just get away with saying, "not our problem" and not trying to get it fixed
c) while i don.t think this fix is right, the basic idea hold s true. whatever the reason for not being able to use the getter/setter, it should fall back to dirty-checking, which from looking at the code it doesn.t

@ccantill you can fix it locally by adding @observable to your decorated proberties or use a fixed decorators implementation `npm i github:doktordirk/babel-plugin-transform-decorators-legacy#release --save-dev`

ccantill:
@jdanyow I agree that the way babel transpiles the decorated props wasn't optimal and that that fixes it, but I still think aurelia should fall back to dirty checking if it can't reconfigure the property rather than just failing silently. I'm sure there could be cases where a non-reconfigurable property needs to be observed without the possibility of changing the code for the property (third party libraries for example).

@doktordirk thanks! I was indeed using `@observable` as a temporary workaround.

jdanyow:
@doktordirk / @ccantill We can't add dirty-checking to fix a transpilation issue because there are many members of the community that are trying to avoid dirty-checking at all costs. They would consider this fix to be a bug.

Issue #301, opened 2016-02-02
background-image broken in latest versions


JeroenVinke:
Something changed in the latest version of Aurelia, breaking the `background-image` style.

```
<div style="background-image: url(http://demos.telerik.com/kendo-ui/content/web/Customers/${CustomerID}.jpg);"></div>
```

gets resolved to

![image](https://cloud.githubusercontent.com/assets/2189477/12764020/cf2e9b04-c9f7-11e5-8273-6f94f419f182.png)

which gives a 404 to "localhost:3000/http"

I have reproduced it here: https://github.com/JeroenVinke/background-image-bug/tree/master/src

I reverted to the previous versions and it worked with those

stoffeastrom:
Could it be [this](https://github.com/z-brooks/binding/commit/a0101c0477954a52b16d8aa33f9d9c9d9b222829#diff-0e7b75111aff8a0702dcf38efb8c5278R82)? I'm not a regexp guy so not sure ;)

EisenbergEffect:
Actually, you shouldn't use `style` with interpolation bindings. It isn't supported cross-browser. Try changing it to `css` instead. Let me know if that works.

jdanyow:
The regex is splitting on the `:` in the url.  cc @z-brooks

z-brooks:
I see. I think that should fix it. New regex only splits if `:` is not followed by `/`.

Gheoan:
In my opinion, a regex is not enough, we need a parser.

JeroenVinke:
@EisenbergEffect 

The `css` attribute is not an option unfortunately due to [this issue](https://github.com/aurelia/templating/issues/229). Thanks @z-brooks and @jdanyow  for the quick response (and the proposed fix).

leye0:
With the regex it fails on any url with a port.

![image](https://cloud.githubusercontent.com/assets/3464994/17784828/ff0d9b32-654b-11e6-8a19-ca8dc3cdea07.png)

PR #483, opened 2016-08-23
fix(StyleObserver): parse complex styles

z-brooks:
Fix for https://github.com/aurelia/templating-resources/issues/248

Should parse anything with function(s), strings or plain values:
_example patterns_
```
abc-xyz: value;  abc: "value"; abc: 'value';
abc: func("..."); abc: func('...'); abc: func(...);
abc: func(test), func("test") func('...') "s:t\"ri;ng: ";
abc: e-f(g(...), xyz); 
background: url(data:image/gif;base64,X//YZ...);
```

### DIFF
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -74,14 +74,15 @@ export class StyleObserver {
           }
         }
       } else if (newValue.length) {
-        let pairs = newValue.split(/(?:;|:(?!\/))\s*/);
-        for (let i = 0, length = pairs.length; i < length; i++) {
-          style = pairs[i].trim();
+        let rx = /\s*([\w\-]+)\s*:\s*((?:(?:[\w\-]+\(\s*(?:"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[\w\-]+\(\s*(?:^"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^\)]*)\),?|[^\)]*)\),?|"(?:\\"|[^"])*"|'(?:\\'|[^'])*'|[^;]*),?\s*)+);?/g;
+        let pair;
+        while( (pair = rx.exec(newValue)) !== null )
+        {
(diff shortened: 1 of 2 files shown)


jdanyow:
now that's what I call a regex :)

Gheoan:
I guess [*a regex*](https://github.com/aurelia/binding/issues/301#issuecomment-178907182) is enough.

stsje:
It will be very difficult to maintain such a long and complex regex.
Would it be possible to have an implementation with less use of regex?

Maybe cssText can be used. A challenge is then removing old unused styles, we have no way of getting those.
https://jsfiddle.net/965qtx73/2/
```
var newValue = 'width: 50px; height: 40px; background-image: url("http://aurelia.io/test2.png"); background-color: blue;';

console.time("Style update");

var element = document.getElementById('element');
var tmpStyleElement = document.createElement('div');

tmpStyleElement.setAttribute('style', element.style.cssText + newValue);
element.setAttribute('style', tmpStyleElement.style.cssText);

console.timeEnd("Style update");
```
Pros are that it's quite simple and performant. Though the implementation is lacking removal of styles.

stsje:
You sure we can't find a more maintainable solution?

jdanyow:
@stsje PRs welcome!

I liked the old way, where we set `style.cssText` but this caused problems when integrating with other libraries that were also changing the element's styles, so we need to make sure the binding doesn't stomp on styles it doesn't "own".

stsje:
@jdanyow What if we changed the way Aurelia supports style & class interpolation?
I don't see a way we can fix all the issues that might come with using a regex, and we certainly don't want to implement a css parser.
Issues like the one I just opened will probably keep popping up from time to time https://github.com/aurelia/templating-resources/issues/251

Currently I don't see a way we can use the native objects such as CSSRuleList to build a solid solution with the current interpolation.

PR #489, opened 2016-09-02
Add property key to change handler arguments

RobertHerhold:
I ran into an issue where I could not determine which property was causing the handler function to be called when using the same function for watching multiple different properties. This PR adds the name of the property as the third argument of the change handler function, which resolves my issue.


jdanyow:
@RobertHerhold nice- thanks!

@EisenbergEffect CLA check

EisenbergEffect:
CLA Confirmed

PR #490, opened 2016-09-02
chore(karma): polyfill tests

Gheoan:
Fixes https://github.com/aurelia/polyfills/issues/33.


jdanyow:
thanks @Gheoan! I thought that module was already loaded but I guess they don't execute in the same order...

Gheoan:
This change should be applied to all repos. Also, I forgot to remove the setup.js imports from specs.

PR #491, opened 2016-09-02
chore(jspm): set transpiler to false

Gheoan:


jdanyow:
@Gheoan what's this one for? Looks like this is in all the repos.

Gheoan:
If the transpile property is missing, jspm install sets the transpiler property to babel (downloading core-js, etc) and that is not needed as karma already transpiles the files.

EisenbergEffect:
@jdanyow I think this is fine to merge in. After doing that can you update the dependency versions for the aurelia deps. Some of them appear to still be on RC.

PR #497, opened 2016-09-08
fix(StyleObserver): enable !important override, remove regex parsing

z-brooks:
No longer using regex to parse the styles. Enables setting a style with `!important` priority to override an existing `!important` style.

aurelia/templating-resources/issues/251

### DIFF
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -53,7 +53,6 @@ export class StyleObserver {
     this.propertyName = propertyName;
 
     this.styles = null;
-    this.version = 0;
   }
 
   getValue() {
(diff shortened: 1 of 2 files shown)


stsje:
We should probably add priority tests by using `style.getPropertyPriority('property name');`.

Iterating the CSSStyleDeclaration obejct comes with a couple of issues.
Setting the background like `element.style.cssText = 'background: url(gaejgioaejgioeaj);';`, will result in 9 style property changes on the object. Instead of ending out with a pretty styling like `background: url(gaejgioaejgioeaj);`, we end up with background-image, background-attachment, background-repeat, ..., background-color all being set?

Gheoan:
Is this affected by #290?

jdanyow:
Definitely an interesting idea. Couple things to check before we can merge:

1. As @Gheoan mentioned, we need to make sure we don't regress on #290. 
2. Performance. Using the current logic as a baseline, how much faster/slower is this? One suggestion- create the div once and reuse it. Here's a gist you can use for benchmarking: https://gist.run/?id=72ba5c3588d6375fee052fc8358a4606

z-brooks:
#290 should be covered. I put [this](https://github.com/z-brooks/binding/blob/0d62c79e3ee661173e823fba13a6764b79e074da/test/element-observation.spec.js#L158) in to test that styles set externally are not affected.

I'll put together a benchmark in a bit.

z-brooks:
Here's the [benchmark](https://gist.run/?id=19d7a99852957195ad5dac3b88a9bb24). Let me know if anything is wrong.  The regex + the `!important` fix is the _fastest_ for me.

jdanyow:
nice work @z-brooks looks like your regex fix actually improved performance too!

Assuming we're in agreement that we need to stick with your original regex approach, what's next? Close this PR and merge #498 ?

z-brooks:
#498 seems the best option. :+1:

PR #498, opened 2016-09-08
fix(StyleObserver): enable !important override

z-brooks:
Inline styles must use `!important` to overwrite `!important` stylesheet rules.

fixes aurelia/templating-resources/issues/251

### DIFF
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -60,6 +60,16 @@ export class StyleObserver {
     return this.element.style.cssText;
   }
 
+  _setProperty(style, value) {
+    let priority = '';
+
+    if (value.indexOf('!important') !== -1) {
+      priority = 'important';
+      value = value.replace('!important', '');
+    }
+    this.element.style.setProperty(style, value, priority);
+  }
+
   setValue(newValue) {
     let styles = this.styles || {};
     let style;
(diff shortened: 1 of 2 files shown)


src/element-observation.js
```diff
@@ -60,6 +60,15 @@ export class StyleObserver {
     return this.element.style.cssText;
   }
 
+  _setProperty(style, value) {
+    let priority = null;
```
Gheoan: `priority` should be empty string (`''`) or `'important'`. [See step 5 from spec](https://www.w3.org/TR/cssom-1/#dom-cssstyledeclaration-setproperty).

z-brooks: hmm, correctly used `''` in the benchmarks :confused:.  It's been changed thanks.

jdanyow:
@z-brooks looking good- one more thing to test:

```shell
karma start --browsers IE
```

I think the `.includes(...)` call might need to be replaced with a regex otherwise the changes won't work in IE.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

z-brooks:
@jdanyow changed it to `.indexOf(...)` and tested in IE.

Gheoan:
Array.prototype.includes is polyfilled by aurelia-polyfills. Maybe tests are not polyfilled corectly again.

Issue #501, opened 2016-09-13
@observable is not a proper decorator


jods4:
https://github.com/aurelia/binding/blob/master/src/decorator-observable.js#L45
The code of `observable` decorator ends with 
`Reflect.defineProperty(target, key, descriptor);`
What it should do is:
`return descriptor;`
The property will be installed by the system, according to the decorators specs.

Why does it matter?
Because installing the property yourself prevents chaining decorators.

For instance, `observable` creates a non-enumerable property with no way to change this.
So I tried to do this:
```ts
@enumerable
@observable
x = 0;
```
Where `@enumerable` simple sets `descriptor.enumerable = true`.

With the way `observable` is coded this doesn't works (`descriptor` is `undefined`).
With my suggestion, it works.

_Aside:_ unlike the property observer, `@observable` really messes up the object on which it is applied (for example if you try to convert it to JSON later). 
Two points specifically:
- Aurelia property observer installs a property that is enumerable (or not) according to the original field/property. Would be nice if `@observable` did the same.
- `observable` installs a backing field simply with convention `_fieldName`. This has side-effects, notably it is serialized in JSON. I understand that this attribute works per class rather than per instance, which makes it more tricky (property observation stores it inside a capture in the getter function, won't work here). At least I think that putting it in a non-enumerable field would be nice. Using a (native) `WeakMap` where available is probably even better.

jdanyow:
@jods4 I agree with everything you wrote, minus the WeakMap part.

Any chance you'd be willing to contribute a PR with the necessary changes?

The trick is getting things to work in both TypeScript and Babel which work slightly differently when applying property decorators. This is maddening, and makes the already complex code even more complex. I put together a little chart, which may need to be updated: https://github.com/aurelia/binding/blob/master/src/decorator-observable.js#L55-L66

jods4:
@jdanyow see PR #503 for the chainable aspect (trivial change to `return decorator`)

About that table, is it still right? 
If I understand it correctly, it implies TS omits the `descriptor` parameter but when I look at the code generated by current TS compiler, it seems to be there.

I think this might have been fixed and TS follows the current standard proposal?

jods4:
@jdanyow I see what you meant now. The table is not obvious ;)
The trick is that _for simple fields_ (not accessors), there is no specs at the moment.
Which means the following code is kind of undefined:
```ts
class C {
  @observable x;
}
```
[...]

PR #503, opened 2016-09-16
Partial fix for #501

jods4:
Makes the decorator chainable by letting the system install the property and return the descriptor.
This doesn't change the observable implementation strategy (enumerable field, non-enumerable accessor, etc.)

### DIFF
--- a/src/decorator-observable.js
+++ b/src/decorator-observable.js
@@ -23,9 +23,11 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
     } else {
       // there is no descriptor if the target was a field, 
       // or if the decorator was applied to a class.
-      descriptor = {
-        enumerable: true  // if the target was a field, it was enumerable
-      };
+      descriptor = { };
+    }
+    // make the accessor enumerable by default, as fields are enumerable
+    if (!('enumerable' in descriptor)) {
+      descriptor.enumerable = true;
     }
 
     // we're adding a getter and setter which means the property descriptor
--- a/src/decorator-observable.js
+++ b/src/decorator-observable.js
@@ -1,7 +1,8 @@
 export function observable(targetOrConfig: any, key: string, descriptor?: PropertyDescriptor) {
   function deco(target, key, descriptor, config) { // eslint-disable-line no-shadow
     // class decorator?
-    if (key === undefined) {
+    const isClassDecorator = key === undefined;
+    if (isClassDecorator) {
       target = target.prototype;
       key = typeof config === 'string' ? config : config.name;
     }
@@ -42,7 +43,12 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
     // dependencies. This is the equivalent of "@computedFrom(...)".
     descriptor.get.dependencies = [innerPropertyName];
 
-    return descriptor;
+    if (isClassDecorator) {
+      Reflect.defineProperty(target, key, descriptor); 
+    }
+    else {
+      return descriptor;
+    }
   }
 
   if (key === undefined) {
--- a/src/decorator-observable.js
+++ b/src/decorator-observable.js
@@ -21,9 +21,9 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
         target[innerPropertyName] = descriptor.initializer();
       }
     } else {
-      // there is no descriptor if the target was a field, 
+      // there is no descriptor if the target was a field in TS (although Babel provides one), 
       // or if the decorator was applied to a class.
-      descriptor = { };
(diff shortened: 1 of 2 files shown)


jdanyow:
@jods4 the circleci build failed- it looks like these tests aren't quite right: https://github.com/aurelia/binding/blob/master/test/decorator-observable.spec.js#L73-L110

Based on the code I see here: https://github.com/aurelia/metadata/blob/master/src/decorators.js#L19

It looks like some args are missing in the `.on(` call.

cc @Gheoan

jods4:
@jdanyow OK this will need more work to get everything green. 
Should I delete this PR? I thought it would be a very quick fix.

At first glance, I think the `decorators` helper is not applying the decorator correctly. Specifically, it only installs the property when passed a `key` but none is passed in this case. It seems `@observable` was doing part of the work for the tests.

jdanyow:
Would you be willing to add another commit to the PR that fixes the tests?

jods4:
@jdanyow not immediately, maybe later. Since there is some work to do, I shall have a look at the enumerability as well.

Gheoan:
From https://github.com/aurelia/metadata/blob/1.0.0/src/decorators.js#L9 it seems that `on` must be called with `key` and `descriptor` only when appling the decorator to a method.

jdanyow:
@Gheoan would you mind fixing up those tests? Then we can merge @jods4's PR right away.

Gheoan:
I don't think it is possible with the current `decorators` implementation.

jods4:
The issue was my change, not the tests or the `decorators` implementation.
I overlooked that you are supposed to be able to use `@observable` in two different ways: either as a property decorator _or as a class decorator (which I overlooked)._

A class decorator should return `undefined` or a new class `Function` (per the specs). It cannot return a descriptor and the system won't define a property for you.

Doing `@observable('prop') class C { }` doesn't feel proper at all but I guess there is some language compat. reason to have it (I guess?).

So my fix was to define the property as before if the decorator was put on a class, or return the descriptor if it was put on a field/accessor.

jods4:
I also make properties enumerable by default (+ corresponding test).
This is to avoid such surprises:
```ts
class X1 {
  value = 42;
}
class X2 {
  @observable value = 42; 
}
JSON.stringify(new X1()); // { "value": 42 }
JSON.stringify(new X2()); // { "_value": 42 } where is 'value'?!
// After enumerable: true
JSON.stringify(new X2()); // { "value": 42, "_value": 42 }
```

jdanyow:
@jods4 re: https://github.com/aurelia/binding/pull/503#issuecomment-247811057

I believe it's to have parity with `@bindable` which works as a property decorator and as a class decorator.

src/decorator-observable.js
```diff
@@ -20,7 +21,13 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
         target[innerPropertyName] = descriptor.initializer();
       }
     } else {
-      descriptor = {};
+      // there is no descriptor if the target was a field, 
+      // or if the decorator was applied to a class.
+      descriptor = { };
```
atsu85: unintended whitespace change?

jods4: yup. and the comment is not 100% accurate.

jods4:
@jdanyow I guess so, but then it doesn't explain clearly why `bindable` can be applied as a class decorator :wink:

doktordirk:
only babels implementation of property decorators is bugged afaik

jods4:
@doktordirk bugged according to what? AFAIK class fields (not accessors) are not standardized, even less decorators on those. So the behavior of `class { @observable x; }` is pretty much undefined.

Turns out TS passes `undefìned` as descriptor in this case, Babel passes an actual descriptor with some reasonable values (enumerable, writable).

doktordirk:
i guess you're right. but maybe then some higherups should enter the discussion about the public fields propsalhttps://github.com/tc39/proposal-class-public-fields/issues/36

jdanyow:
right

jods4:
@jdanyow the failing CI check is a timeout on closing Chrome, not sure why / what I can do. It was green on the second-to-last commit, and the last commit is just comments/whitespace changes.

jdanyow:
@jods4 cool- so this is ready to merge?

jods4:
@jdanyow yes for enumerable properties and chainable decorators.
The part about hiding `_name` backing field is not done, though, because I am unsure about how to tackle this.

jdanyow:
thanks @jods4, I opened https://github.com/aurelia/binding/issues/505 to track the remaining work

Issue #505, opened 2016-09-23
observable decorator creates backing field that shouldn't be enumerable


jdanyow:
This issue was split off from https://github.com/aurelia/binding/issues/501

cc @jods4

Gheoan:
Even if the backing property is converted to accesors, it can still be nonenumerable.

jods4:
The problem with the backing property is that it must be observable by Aurelia's regular observer.
So it must be a regular property or we need a custom observation strategy (e.g. add `getObserver` on the getter) on the property decorated with `@observable`.
If it is a regular property, it can still be non-enumerable, but then you have to use `Object.defineProperty('_name', { writable: true, enumerable: false, value: 42 })` _each time you set its value_.
And after Aurelia `SetterObserver` has transformed this plain property into accessors to be able to observe it, the call above won't work anymore.

EisenbergEffect:
@jdanyow Where are we at with this?

jdanyow:
The PR is still in review

PR #506, opened 2016-09-24
fix(observable): inner property should be non-enumerable

Gheoan:
Fix https://github.com/aurelia/binding/issues/505.

### DIFF
--- a/src/decorator-observable.js
+++ b/src/decorator-observable.js
@@ -9,6 +9,11 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
 
     // use a convention to compute the inner property name
     let innerPropertyName = `_${key}`;
+    const innerPropertyDescriptor: PropertyDescriptor = {
+      configurable: true,
+      enumerable: false,
+      writable: true,
+    };
 
     // determine callback name based on config or convention.
     const callbackName = (config && config.changeHandler) || `${key}Changed`;
@@ -18,7 +23,7 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
 
       // set the initial value of the property if it is defined.
       if (typeof descriptor.initializer === 'function') {
-        target[innerPropertyName] = descriptor.initializer();
+        innerPropertyDescriptor.value = descriptor.initializer();
       }
     } else {
       // there is no descriptor if the target was a field in TS (although Babel provides one),
@@ -35,11 +40,18 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
     delete descriptor.writable;
     delete descriptor.initializer;
 
+    // Add the inner property on the prototype.
+    Reflect.defineProperty(target, innerPropertyName, innerPropertyDescriptor);
+
     // add the getter and setter to the property descriptor.
     descriptor.get = function() { return this[innerPropertyName]; };
     descriptor.set = function(newValue) {
       let oldValue = this[innerPropertyName];
+
+      // Add the inner property on the instance and make it nonenumerable.
       this[innerPropertyName] = newValue;
+      Reflect.defineProperty(this, innerPropertyName, { enumerable: false });
+
       if (this[callbackName]) {
         this[callbackName](newValue, oldValue, key);
       }
--- a/test/decorator-observable.spec.js
+++ b/test/decorator-observable.spec.js
@@ -1,6 +1,8 @@
 import './setup';
 import {observable} from '../src/decorator-observable.js';
 import {decorators} from 'aurelia-metadata';
+import {SetterObserver} from '../src/property-observation';
+import {Logger} from 'aurelia-logging';
 
 describe('observable decorator', () => {
   const oldValue = 'old';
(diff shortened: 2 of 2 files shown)


Gheoan:
Why is the inner property on the prototype instead of the instance?

jdanyow:
@Gheoan nice! This looks good to me.

@jods4, thoughts?

jods4:
@Gheoan because `target` is the prototype of the class for non-static members, that is expect.

I have not run the code, but I think it doesn't work. 
Please add a test that checks if the backing field is still non-enumerable _after_ its value has been changed by client code.

test/decorator-observable.spec.js
```diff
@@ -108,15 +112,41 @@ describe('observable decorator', () => {
...
-      keys.push(p);
-    }
-    expect(keys).toContain('value');
+
+    expect(Object.getPrototypeOf(instance).propertyIsEnumerable('value')).toBe(true);
+  });
+
+  describe('inner property', () => {
+    it('should exist', () => {
+      const instance = new class {
```
atsu85: Why is the const called `instance` when controversially it seems to be a class, not instance of the class?

doktordirk:
has that been tried with babel?

test/decorator-observable.spec.js
Gheoan: It is an instance. Note the `new` keyword.

jdanyow: right- anonymous class

jdanyow:
@Gheoan can you add the test @jods4 requested, assuming it passes we can merge.

Thank you both for all of your work on this!

test/decorator-observable.spec.js
Gheoan: It is an instance. Note the `new` keyword.

Gheoan:
Added more tests. @jods4 let me know if this looks good.

jods4:
@Gheoan I am not sure the `Object.propertyIsEnumerable()` test is good enough. It does always return `false` for properties inherited from prototypes. 

When I was prototyping what you've done, I was worried because of this code:
```javascript
function X() {}
Object.defineProperty(X.prototype, 'y', { configurable: true, enumerable: false, writable: true });
var x = new X();
x.y = 42;
Object.propertyIsEnumerable(x, 'y'); // -> false
for (k in x) console.log(x); // -> 'y'
JSON.stringify(x);  // -> { "y": 42 }
```

Gheoan:
@jods4 In the example from your [comment](https://github.com/aurelia/binding/pull/506#issuecomment-250417982): 
- on line 2 a **non-enumerable** property is created on the prototype of `Function` object `X` (capital).
- on line 4 a property is created on the object `x`. Properties created this way are enumerable by default (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties).
- line 5 is not quite right. The rigth way would be `x.propertyIsEnumerable('y')` which is `true` because of line 2.

I will check to be sure that `SetterObserver` does create the accesors on the right object to prevent what happens on line 2 of your example.

Gheoan:
Changed tests to reflect my comment above.

test/decorator-observable.spec.js
```diff
@@ -108,15 +111,69 @@ describe('observable decorator', () => {
...
+        '_value'
+      );
+
+      it('should convert', () => {
+        spyOn(Logger.prototype, 'warn');
+        observer.convertProperty();
+        expect(Logger.prototype.warn).not.toHaveBeenCalled();
+      });
+
+      it('should exist', () => {
```
Gheoan: This test asserts that `SetterObserver` creates the getter-setter on the right object (https://github.com/aurelia/binding/pull/506#issuecomment-250547363).

jods4:
@Gheoan but isn't it exactly what happens in the decorator code?
  ```ts
Reflect.defineProperty(target, innerPropertyName, {
  configurable: true,
  enumerable: false,
  writable: true
});
```
Here `target` is the prototype of the class, right?
```ts
descriptor.set = function(newValue) {
  let oldValue = this[innerPropertyName];
  this[innerPropertyName] = newValue;
  if (this[callbackName]) {
    this[callbackName](newValue, oldValue, key);
  }
};
```
And here, the first time the value is set, `this[innerPropertyName]` will create an enumerable property directly on the instance?

Gheoan:
Right, I changed the setter to use `target` instead of `this` and added a test.

jdanyow:
@Gheoan `target` cannot be used can it? Wouldn't that make all instances of the class have the same value because target is the prototype?

src/decorator-observable.js
```diff
@@ -36,10 +42,10 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
     delete descriptor.initializer;
 
     // add the getter and setter to the property descriptor.
-    descriptor.get = function() { return this[innerPropertyName]; };
+    descriptor.get = function() { return target[innerPropertyName]; };
     descriptor.set = function(newValue) {
-      let oldValue = this[innerPropertyName];
-      this[innerPropertyName] = newValue;
+      let oldValue = target[innerPropertyName];
+      target[innerPropertyName] = newValue;
```
jdanyow: @Gheoan I don't think you can use `target` in this way. It creates a closure and is assigning the value to the prototype. I think if you write a test where you have two instances of an `@observable` decorated class, and you set the property on one instance, the second instance will have the same property value.

Gheoan:
Did some changes. The private property is added first to the prototype, then to the instance. It is nonenumerable in both cases.

jods4:
@Gheoan Clever! JS is so tricky sometimes... so you set the property value normally so it works on both direct access and getter/setters, and then set _only_ the `enumerable` flag to false to hide it. I guess that works in all cases, good idea.

src/decorator-observable.js
```diff
@@ -9,6 +9,11 @@ export function observable(targetOrConfig: any, key: string, descriptor?: Proper
 
     // use a convention to compute the inner property name
     let innerPropertyName = `_${key}`;
+    const innerPropertyDescriptor: PropertyDescriptor = {
+      configurable: true,
+      enumerable: false,
+      writable: true,
+    };
```
jods4: Is this still useful with the new implementation of the setter? What happens if you remove it?

Gheoan: Do you mean not creating the inner property? Doing so causes some test that were not added in this PR to fail.

jods4: OK. Maybe they should be included or documented then? It's not obvious why this is still required (I could probably come up with at least one funny case involving `with` binding or overriden contextes).

Gheoan: Sorry for my confusing comment above. Removing [this](https://github.com/aurelia/binding/pull/506/files#diff-26d3f601c99bc9cd08a30349643a0ac2R44) line causes some old, already existing test to fail.

Gheoan: Sorry for my confusing comment above. Removing [this](https://github.com/aurelia/binding/pull/506/files#diff-26d3f601c99bc9cd08a30349643a0ac2R44) line causes some old, already existing test to fail.

jdanyow:
@Gheoan thanks for all the hard work 👍

PR #513, opened 2016-09-29
Update to typings file in dist

jeppevammenkristensen:
Fixing an issue where the compiler for typescript reported an exception for the binding typings file


jdanyow:
this is fixed in the latest build

Issue #511, opened 2016-09-29
observable does not define property since 1.0.5


Thanood:
**I'm submitting a bug report**

* **Library Version:**
1.0.5

**Please tell us about your environment:**
* **Operating System:**
Windows 8

* **Node Version:**
6.2.2

* **NPM Version:**
3.10.5

* **JSPM OR Webpack AND Version**
JSPM 0.16.15

* **Browser:**
all

* **Language:**
all

**Current behavior:**

`observable` does not define the given property on the target object unless it's a class decorator.
This behaviour was introduced in https://github.com/aurelia/binding/commit/6586cd2 (line 52).

**Expected/desired behavior:**

* **What is the expected behavior?**

If not present, the property is defined in the target object.

* **What is the motivation / use case for changing the behavior?**

I have to use ES5 in parts of my current app. I'm defining observables like so:

```javascript
var fw = require('aurelia-framework');
function MyClass() {
  fw.observable(this, 'value');
}
MyClass.prototype.attached() {
  this.value = false;
}
MyClass.prototype.valueChanged(newValue) {
  fireEvent(this.checkbox, 'change', { value: newValue });
}
```

Because the `value` property is not defined in the constructor (by `observable`) with a setter decoration, it is defined (overwritten) on attached. The `valueChanged` function is never called. This broke all of my filters after updating from v1.0.4. :smiley:

Edit: it doesn't help to define `value` before calling `observable`.
Downgrading to 1.0.4 fixed this for me.

estaphorsius:
I'm having the same issue over here. We are using aurelia with TypeScript. Building our application results in a compilation error:

Error: typings/modules/aurelia-binding/index.d.ts(193,14): error TS2420: Class 'DataAttributeObserver' incorrectly implements interface 'InternalPropertyObserver'.
  Property 'getValue' is missing in type 'DataAttributeObserver'.

It seems typing for DataAttributeObserver is missing a definition.

fracz:
I can confirm it is fixed in 1.0.6. Thank you!

Gheoan:
@estaphorsius, @fracz I think this is a different issue.
@Thanood, it works for me: https://gist.run/?id=b1f2df7ae68e97d47bb3a7e7d696cebe.

Thanood:
@Gheoan Depends on the version used in the gist's bundle.
I tried a search for `isClassDecorator` on the bundle and it had no search results. That could mean it's still at 1.0.4.

jdanyow:
[...]

PR #514, opened 2016-09-29
fix(observable): fix descriptor having both value and get and set

Gheoan:
Related: https://github.com/aurelia/binding/issues/511.

### DIFF
--- a/test/decorator-observable.spec.js
+++ b/test/decorator-observable.spec.js
@@ -70,43 +70,63 @@ describe('observable decorator', () => {
     expect(instance.value).toEqual(newValue);
   });
 
-  it('should work with decorators function', () => {
-    const instance = new (decorators(observable('value'))
-      .on(class {
+  describe('es2015 with decorators function', () => {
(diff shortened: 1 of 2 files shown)


Gheoan:
Added a new test and changed old ones.

test/decorator-observable.spec.js
```diff
@@ -70,45 +70,76 @@ describe('observable decorator', () => {
     expect(instance.value).toEqual(newValue);
   });
 
-  it('should work with decorators function', () => {
-    const instance = new (decorators(observable('value'))
-      .on(class {
+  describe('es2015 with decorators function', () => {
+    xit('should work when decorating property', () => {
```
Gheoan: This test is pending https://github.com/aurelia/metadata/pull/44.

Gheoan: No longer pending.

EisenbergEffect:
@jdanyow Assigned back to you since @Gheoan appears to have added the test :)

PR #522, opened 2016-10-06
fix(styleobserver): improper value check in if

Vheissu:
The previous if statement check would break things if the value was undefined, as such, indexOf which was being checked for would not exist on an undefined/null value. The better check to do is just ensure that value exists as this seems to be a better way of handling this numeric edge case.

### DIFF
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -63,7 +63,7 @@ export class StyleObserver {
   _setProperty(style, value) {
     let priority = '';
 
-    if (value.indexOf && value.indexOf('!important') !== -1) {
+    if (value && value.indexOf('!important') !== -1) {
       priority = 'important';
       value = value.replace('!important', '');
     }
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -63,7 +63,7 @@ export class StyleObserver {
   _setProperty(style, value) {
     let priority = '';
 
-    if (value && value.indexOf('!important') !== -1) {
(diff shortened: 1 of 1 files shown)


EisenbergEffect:
I'm not super familiar with this part of the code. Will your fix work if the there's a string that is falsey?

Vheissu:
Good pickup @EisenbergEffect I actually just made this a little more bulletproof. I first check if the value is of type undefined (so null, etc) and then I also check if the .indexOf property is defined using a typeof check as well (numeric values don't have a indexOf property). The previous commit would have indeed failed for numeric values which were 0 and other potential catches.

src/element-observation.js
```diff
@@ -63,7 +63,7 @@ export class StyleObserver {
   _setProperty(style, value) {
     let priority = '';
 
-    if (value && value.indexOf('!important') !== -1) {
+    if (typeof value !== typeof undefined && typeof value.indexOf !== typeof undefined && value.indexOf('!important') !== -1) {
```
EisenbergEffect: It could still be null. I think we want something like this:

```
if (value !== null && value !== undefined && typeof value.indexOf === 'function' && value.indexOf('!important') !== -1)
```

Does that make sense? If so can you update the PR with that?

Vheissu: That makes sense and it still works for me, in the midst of rushing to get this fixed, I completely didn't realise that I wasn't doing a proper null check there. I'll update it with that code.

PR #524, opened 2016-10-08
fix(StyleObserver): convert camelCase properties to kebab-case

### LINKED ISSUE #523: Binding style object works only on single worded css properties [1.0.9]
I have an element with css.bind="style".
In Viewmodel I have:
```
	get style() {
		return {
			backgroundColor: blabla,
			color: blabla,
			padding: blabla,
			fontFamily: blabla,
			fontWeight: blabla,
			fontStyle: blabla,
			fontSize: blabla,
			textAlign: blabla
		};
	}
```

But only color and padding gets binded. I just assume its because other properties are not single words.
Previously I used version 1.0.4 and everything worked fine.

z-brooks:
The property methods (`set|removeProperty`) don't recognize _camelCased_ names.

fixes #523

### DIFF
--- a/src/element-observation.js
+++ b/src/element-observation.js
@@ -77,10 +77,13 @@ export class StyleObserver {
 
     if (newValue !== null && newValue !== undefined) {
       if (newValue instanceof Object) {
+        let value;
         for (style in newValue) {
           if (newValue.hasOwnProperty(style)) {
+            value = newValue[style];
+            style = style.replace(/([A-Z])/g, m => '-' + m.toLowerCase());
             styles[style] = version;
-            this._setProperty(style, newValue[style]);
+            this._setProperty(style, value);
           }
         }
       } else if (newValue.length) {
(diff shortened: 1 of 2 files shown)


src/element-observation.js
```diff
@@ -70,17 +70,24 @@ export class StyleObserver {
...
     let style;
     let version = this.version;
 
     if (newValue !== null && newValue !== undefined) {
       if (newValue instanceof Object) {
+        let value;
         for (style in newValue) {
           if (newValue.hasOwnProperty(style)) {
+            value = newValue[style];
+            style = style.replace(/([A-Z])/g, this._fixCase);
```
jdanyow: @z-brooks what do you think about changing this to `.replace(...., m => '-' + m.toLowerCase());` instead of using the class method?

z-brooks: @jdanyow  Yeah, that's fine. I had originally done that but I wasn't sure and thought closures and the like were discouraged.

jdanyow: thanks for changing that- arrow functions are ok- it's only a closure if it accesses the outer scope.

PR #530, opened 2016-10-31
feat(EventManager): Support capturing event

bigopon:
@jdanyow @EisenbergEffect 
This is just a seed based on Jeremy's prototype [Gistrun](https://gist.run/?id=196c99652a28d571527a8fff63297308)

* **Purpose**: Provide a way to listen to capture event

* **Note**: this needs to come together with `SyntaxInterpreter` class changed. Which depends on probablility of this pr.

I'm not sure if this is in Aurelia team's plan so I only committed this. If ok, then I'll update all other repos where `ListenerExpression` is used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-binding",
-  "version": "1.0.9",
+  "version": "1.1.0",
   "description": "A modern databinding library for JavaScript and HTML.",
   "keywords": [
     "aurelia",
--- a/package.json
+++ b/package.json
@@ -21,10 +21,17 @@
   },
   "jspm": {
     "registry": "npm",
+    "jspmPackage": true,
     "main": "aurelia-binding",
     "format": "amd",
     "directories": {
       "dist": "dist/amd"
+    },
+    "peerDependencies": {
+      "aurelia-logging": "^1.0.0",
+      "aurelia-metadata": "^1.0.0",
+      "aurelia-pal": "^1.0.0",
+      "aurelia-task-queue": "^1.0.0"
     },
     "dependencies": {
       "aurelia-logging": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "aurelia-binding",
-  "version": "1.1.0",
+  "version": "1.0.9",
   "description": "A modern databinding library for JavaScript and HTML.",
   "keywords": [
     "aurelia",
@@ -21,17 +21,10 @@
   },
   "jspm": {
     "registry": "npm",
-    "jspmPackage": true,
     "main": "aurelia-binding",
     "format": "amd",
     "directories": {
       "dist": "dist/amd"
-    },
-    "peerDependencies": {
-      "aurelia-logging": "^1.0.0",
-      "aurelia-metadata": "^1.0.0",
-      "aurelia-pal": "^1.0.0",
-      "aurelia-task-queue": "^1.0.0"
     },
     "dependencies": {
       "aurelia-logging": "^1.0.0",
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -72,6 +72,25 @@ export declare class BindingBehaviorResource {
  */
 export declare function subscriberCollection(): any;
 
+
+/**
+ * Describes the strategy phase where event should be handled.
+ */
+export declare enum eventHandlingMethod {
+  /**
+   * Default strategy, event listener is bound to the listening element
+   */
+  normal = 0,
+  /**
+   * Capture strategy, listening on root level during event capturing phase
+   */
+  capture = 1,
+  /**
+   * Delegate strategy, listening on root level during event bubbling phase
+   */
+  delegate = 2
+}
+
 /**
  * Subscribes to appropriate element events based on the element property
  * being observed for changes.
@@ -87,7 +106,7 @@ export declare class EventManager {
    * @param delegate True to use event delegation mechanism.
    * @returns function wich removes event listener.
    */
-  addEventListener(target: Element, targetEvent: string, callback: (event: Event) => any, delegate: boolean): () => void;
+  addEventListener(target: Element, targetEvent: string, callback: (event: Event) => any, delegate: eventHandlingMethod): () => void;
 }
 
 /**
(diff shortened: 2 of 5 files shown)


bigopon:
Clearly I didn't read the contribution guidelines. If this looks good to go ahead, I'll edit the commit message

jdanyow:
I think this would be a nice addition. @EisenbergEffect - thoughts?

@bigopon what do you think about changing `eventHandlingMethod` from:

```js
const eventHandlingMethod = {
   /**
    * Default strategy, event listener is bound to the listening element
    */
   normal: 0,
   /**
    * Capture strategy, listening on root level during event capturing phase
    */
   capture: 1,
   /**
    * Delegate strategy, listening on root level during event bubbling phase
    */
   delegate: 2
 }
```

to:
```js
const delegationStrategy = {
   /**
    * No event delegation.
    */
   none: 0,
   /**
    * Capturing phase event delegation strategy.
    */
   capturing: 1,
   /**
    * Bubbling phase event delegation strategy.
    */
   bubbling: 2
 }
```

EisenbergEffect:
This seems like a nice idea provided we don't break the existing behaviors/syntax. I didn't review the full code, just looked quickly at the gist and the ideas. Also, I like the change that @jdanyow recommends for the enum. If @jdanyow is good with it, let's move ahead.

jdanyow:
Awesome! @bigopon let's proceed 🚀 

If there's anything I can help out with, let me know. Thanks!

bigopon:
Cool. And agree on the naming, in fact I tried `eventStrategy`, `eventBubblingPhase`, `handlingStrategy`, `eventDelegateStrategy` but failed to come up with `delegationStrategy` ....

@jdanyow one thing concerns me is that every time we call the handler, a new array is created to store the ordered callbacks. Do you have any better ideas ?

EisenbergEffect:
That's a real concern and if there's an idea to prevent that, it would be nice...

jdanyow:
@bigopon check the history on that region of the code- any issues tied to that change?

bigopon:
@jdanyow  I don't understand what you meant so I'll explain why I needed to put the array is that event propagates down from parent to child during capturing, we need to store all callbacks for event path and execute them in parent -> child order.

And it's 1am here so I'll try to response asap after this message.

bigopon:
@jdanyow 
Just finished test for this. Gonna update other repos.

jdanyow:
@bigopon please exclude the package.json from the PR. Typically we only include the `src` and `test` artifacts. The version bump happens as part of the release.

bigopon:
@jdanyow Done.

I also committed in 
* `templating-binding` : aurelia/templating-binding#99
* `templating-resources`: aurelia/templating-resources#259

Can you have a look. Sorry I don't know to wire them.

jdanyow:
Excellent- this is looking good. Are you done with the changes? If so, I will pull this down and do a full review in the next few days.

Also- I think I misunderstood your comment here: https://github.com/aurelia/binding/pull/530#issuecomment-257538698

Would you mind linking to the line(s) of code you're referring to?

Thanks!

bigopon:
Yeah, all changes are done and tests are updated accordingly. Just not wired up so some tests fail because they refers to old version.

I was talking about these line [L21-L37](https://github.com/bigopon/binding/blob/3c4568ba1f1e9f4463f3a7fd350fd1d640ead558/src/event-manager.js#L21-L37)

jdanyow:
@bigopon sorry for the delay- I finally had a chance to review this. It's looking really good. Regarding the array allocation, I don't think it will be an issue. We can benchmark it after this is merged. Your changes certainly won't degrade perf of existing apps.

Left you one comment above about a variable declaration, once that's resolved I'll get this merged.

Thanks!

bigopon:
Done. I also fixed the starting index in line 38.

bigopon:
Oh that's bad. Forgot to change after changed the constants 😭 

@jdanyow 
I have one other query related to events, these line [L10 - L13](https://github.com/aurelia/binding/blob/master/src/event-manager.js#L10-L13)

Do we have to create a function for each event ? can we just cache it an assign to every event instead ? It's micro thing but could be good.

jdanyow:
good call- want to send a PR?

bigopon:
Does this look good ? 

#539

gregoryagu:
I would like to use event capturing during drag and drop operations.  

During the event, I need to know the event the event is attached to in the HTML.  

Currently, capture attaches to the document (like event.delegate) instead of the source element(like event.trigger)  

So in the event, when I check the currentTarget, the property is set to the root document, not the element.  So I have no way to know which element is the currentTarget.

If event.capture worked more like event.trigger, that would work for me.

bigopon:
I'm guessing the element listening to drag & drop has descendants so you cannot use `event.target` ? You can use `self` binding behavior for this. But correct me if I'm wrong as I don't know your use case.

gregoryagu:
Yes, there are descendants of the dropzone.  The issue is that the when an item is dragged over a descendant, it cause dragLeave to be fired on the parent (when in fact, the cursor is still within the parent, so it should not be). So the parent loses the highlighting.

So what I wanted to do was capture the dragOver in the parent, and stop the propagation down to the child.  This (hopefully) would prevent the DragLeave from being called on the parent. 

But in any case, I found a [workaround ](http://stackoverflow.com/questions/7110353/html5-dragleave-fired-when-hovering-a-child-element/38416886#38416886)on StackOverflow, and it's working fine.

bigopon:
@gregoryagu 
Hmm i remember hitting this annoyance behavior long ago. Maybe you can disable all pointer event on `dragenter`, it ensures there's nothing will fire `dragleave` within the element

gregoryagu:
@bigopon This looks great.  I tried the gist, and yes it does work.  It doesn't look like it is actually using the "self" class.  I forked the gist and tried without it, and it seems to work fine.  Yes, I was using dragCount.  I would much rather use your solution.  In a few hours, I will have a chance to try it on the actual project and will let you know.  I appreciate the help very much.  The finer points of Aurelia sometimes eludes me!

bigopon:
@gregoryagu this is gonna be your wtf moment i guess 😄, if your inner elements are close to the edge of the container and user drag too fast and there is no `self` binding behavior, the order of the events will be like this: (from the example)
```
Entering: button
Leaving: button
Leaving: header
```

It's because when you drag too fast and enter the button, the header will be added a class to make button `pointer-events: none`, thus triggers the `dragleave` and bubbles up. This is where `self` is needed.

gregoryagu:
@bigopon Interesting.  Yes, I tried it moving the mouse quickly and I did see the effect that self has on it. Awesome.  I just read the doc you wrote on it.  Very interesting.  Good work!  Oh those mouse events!!

PR #532, opened 2016-10-31
fix(typings): fix AccessMember constructor parameters order

valery-vitko:


jdanyow:
thanks @valery-vitko !

PR #539, opened 2016-11-13
fix(EventManager): cache interceptor

bigopon:


src/event-manager.js
```diff
@@ -5,12 +5,14 @@ function findOriginalEventTarget(event) {
   return (event.path && event.path[0]) || (event.deepPath && event.deepPath[0]) || event.target;
 }
 
+function interceptor() {
```
jdanyow: what do you think about naming this function `stopPropagation`?

bigopon:
Done. I dont know how to change it directly here so another commit though

PR #545, opened 2016-12-04
Added missing bracket for input element

akshayKhot:
<input type="text" value.bind="taskName" line has a missing bracket, messing with code highlighting in docs.


jdanyow:
nice- thanks!

PR #550, opened 2016-12-13
fix(map-observation): change "add" property to "set"

### LINKED ISSUE #549: Map observation- "add" should be "set"
https://github.com/aurelia/binding/blob/master/src/map-observation.js#L34-L39

from @ChibiBlasphem
https://github.com/aurelia/binding/issues/155#issuecomment-266259151

ChibiBlasphem:
fixes #549

### DIFF
--- a/src/map-observation.js
+++ b/src/map-observation.js
@@ -31,9 +31,9 @@ class ModifyMapObserver extends ModifyCollectionObserver {
     let observer = new ModifyMapObserver(taskQueue, map);
 
     let proto = mapProto;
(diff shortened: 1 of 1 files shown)


jdanyow:
hey @ChibiBlasphem - thanks for reporting this and putting together the PR. Could you please amend this to exclude the `dist/*` files? Thanks!

ChibiBlasphem:
I think it's good now. I was hesitating on whether to push them or not. Sorry.

jdanyow:
perfect- thanks!

PR #569, opened 2017-02-04
feat(build): make SVG support optional

jods4:
The list of SVG elements and attributes weights around 20K, does not minify and is seldom used.

I put it behind a free variable `FEATURE_NO_SVG`.
If you do nothing special that global is undefined an nothing changes.

The idea here is that tools like webpack can define a free variable `FEATURE_NO_SVG` and then remove all that dead code when minifying.
If you do so, SVG bindings don't work anymore, but your build is around 20K smaller.


EisenbergEffect:
@jdanyow Can you review this please?

PR #573, opened 2017-02-13
refactor(size): Unparser opt-out

jods4:
Add a global free variable `FEATURE_NO_UNPARSER` that can be used to remove `Unparser` from the build.

`Unparser` is mainly a debugging feature that is not used by consumers of Aurelia. The only visible change when removing it is that calling `toString()` on an `Expression` returns `[object Object]` instead printing the AST back to a string.

Removing `Unparser` cuts approx. 2K of minified code.

**Caution**: `Unparser` is presently used by `aurelia-validation` and can't be removed from an application that uses it. See aurelia/validation#412.

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -16,7 +16,9 @@ export class Expression {
   }
 
   toString() {
-    return Unparser.unparse(this);
+    return typeof FEATURE_NO_UNPARSER === 'undefined' ? 
+      Unparser.unparse(this) :
+      super.toString();
   }
 }
 
--- a/src/unparser.js
+++ b/src/unparser.js
@@ -1,189 +1,193 @@
-export class Unparser {
-  constructor(buffer) {
-    this.buffer = buffer;
-  }
+export let Unparser = null;
 
-  static unparse(expression) {
-    let buffer = [];
-    let visitor = new Unparser(buffer);
-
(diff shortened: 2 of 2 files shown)


src/ast.js
```diff
@@ -16,7 +16,9 @@ export class Expression {
   }
 
   toString() {
-    return Unparser.unparse(this);
+    return typeof FEATURE_NO_UNPARSER === 'undefined' ?
```
jdanyow: @jods4 sorry for the delay. This all looks good. If you want to save a few more bits, go ahead and use `return FEATURE_NO_PARSER === undefined ? ... : ...`. We typically don't use the typeof style check for undefined. I think a few contributors have in the past but most of the time we use `something === undefined`. If some jerk redefined what undefined's value is they'll have to deal with it 😉

jods4: This is not the problem. Reading a non-existent variable is an error in strict mode.
`FEATURE_NO_UNPARSER === 'undefined'` will throw if `FEATURE_NO_UNPARSER` does not exist on `window`, which is the common case.

The direct comparison to `undefined` is only safe on existing objects, existing vars or parameters.

jdanyow: ah- that makes sense

PR #580, opened 2017-02-26
feat(observable): typing `PropertyObserver` changed to be `PropertyObserver<T>`

stephenlautier:
changed also the following respectively
- `BindingEngine.propertyObserver<T>`
- `BindingEngine.expressionObserver<T>`

The reason is to provide better typings for the consumers, as currently when using the `PropertyObserver` and exposing it, it will be `any`, now instead it will provide the correct typing.

### DIFF
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -112,11 +112,11 @@ export declare class EventManager {
 /**
  * Observes property changes.
  */
-export declare interface PropertyObserver {
+export declare interface PropertyObserver<T> {
   /**
    * Subscribe to property change events.
    */
-  subscribe(callback: (newValue: any, oldValue: any) => void): Disposable;
+  subscribe(callback: (newValue: T, oldValue: T) => void): Disposable;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


StrahilKazlachev:
Wouldn't this break the compilation of existing apps with errors like `Property 'some_property_name' does not exist on type '{}'.`?

stephenlautier:
@StrahilKazlachev unfortunately, it is breaking; its a bit better error message tho e.g.

`Generic type `PropertyObserver<T>` requires 1 type argument(s).`

When types are not 100% correct and changed unfortunately they are "breaking", but obviously it will be stricter and might also help you fix several issues

StrahilKazlachev:
I'll be honest, I wouldn't introduce a breaking change and still have the same problem - incorrect typings.
If it could be typed like `propertyObserver<T, K extends keyof T>(obj: T, propertyName: K): PropertyObserver<T[K]>;` it would be something else. But this requires TS >=2.1 and I'm not sure if/how this would reflect on the API doc generation.
@jdanyow What do you think?

jdanyow:
let's revisit when we convert the source of this repo to TypeScript

stephenlautier:
@StrahilKazlachev I understand that it introduces a breaking change for typescript users, however as a typescript user I would want it to be typed correctly and not having all "observables" as `any`.

I was also thinking to use the `keyof`, however it does have a problem.. if the property is private it won't work and thats a bigger issue.

It would be nicer if works as `propertyObserver(this, x => x.activeItem)` but the private will also be an issue

EisenbergEffect:
Closing since this should be addressed after a conversion to TypeScript.

PR #600, opened 2017-07-07
feat(binding-expression): add improved error message to updateSource …

peteigel:
…and updateTarget

Added custom error messages to BindingExpression when updateSource() or updateTarget() fails.
These include information about what the source/target that was being set,
and what value it was being set to. Previously the underlying error was allowed to bubble to the top,
and valuable information about the context of the binding expression was not reported.

### DIFF
--- a/src/binding-expression.js
+++ b/src/binding-expression.js
@@ -40,11 +40,21 @@ export class Binding {
   }
 
   updateTarget(value) {
-    this.targetObserver.setValue(value, this.target, this.targetProperty);
+    try {
+      this.targetObserver.setValue(value, this.target, this.targetProperty);
+    } catch (e) {
+      throw new Error(
+        `aurelia-binding: Error encountered updating property ${this.targetProperty} of target ${this.target} to ${value} - ${e}`
+      );
+    }
   }
 
   updateSource(value) {
-    this.sourceExpression.assign(this.source, value, this.lookupFunctions);
+    try {
+      this.sourceExpression.assign(this.source, value, this.lookupFunctions);
+    } catch (e) {
+      throw new Error(`aurelia-binding: Error encountered setting value of source ${source} to ${value} - ${e}`);
+    }
   }
 
   call(context, newValue, oldValue) {


peteigel:
This was submitted after my team spent a large amount of time tracking down a browser-compatibility bug. If updating a property a bound fails (for instance it is read only, or has a custom setter), the error message emitted did not contain any information about the binding expression itself. Likewise, the stack trace is very opaque, usually with several successive calls to `bind()`.

I am not sure if the error messages I wrote are consistent with others in the system. Please let me know what I can do to improve them.

dist/amd/aurelia-binding.js
```diff
@@ -47,7 +47,7 @@ define(['exports', 'aurelia-logging', 'aurelia-pal', 'aurelia-task-queue', 'aure
   var _typeof = typeof Symbol === "function" && typeof Symbol.iterator === "symbol" ? function (obj) {
     return typeof obj;
   } : function (obj) {
-    return obj && typeof Symbol === "function" && obj.constructor === Symbol ? "symbol" : typeof obj;
```
atsu85: please revert the updates to generated files (changes under dist are committed during release process AFAIK)

atsu85: @EisenbergEffect, adding files under dist to the PR is a recurring problem. Perhaps this check could be automated, for example using custom client-side git hook (that could be installed with npm postinstall script)

peteigel: Done. Sorry, this is my first PR for Aurelia. I misunderstood the line in the [Contributor's Guide](https://github.com/DurandalProject/about/blob/master/CONTRIBUTING.md)

> Build your changes locally to ensure all the tests pass

to mean commit up the built files.

jdanyow:
@peteigel thanks for the PR, it makes complete sense. Unfortunately we cannot add try-catch here for performance reasons. It's a really hot code path.

In the future when debugging something like this what you can do is configure your debugger to break on all exceptions and enter `this.sourceExpression.toString()` to view the binding expression that is causing the problem.

PR #610, opened 2017-07-29
Destructor dirty checking

JeroenVinke:


jdanyow:
what is this for

JeroenVinke:
This is necessary for SSR. Because of the setTimeout of the DirtyChecker Node.js won't garbage collect the application, resulting in memory leaks. So in order to fix this we need a way to stop the DirtyChecker.

jdanyow:
Got it- would it make sense to disable the dirty checker entirely in the SSR case? Never even start the interval?

JeroenVinke:
That would be great. It would still do an initial bind, right?

Do you have an idea of the API? We might have to do similar things in other parts of the framework, although I haven't encountered them yet.

jdanyow:
Right, it would still do the initial bind. You could make `scheduleDirtyCheck` a no-op if the platform is node. Probably would need to make [`enqueueBindingConnect`](https://github.com/aurelia/binding/blob/master/src/connect-queue.js) a no-op as well.

This has got me thinking though- I wonder if we should force binding mode to always be `one-time` in SSR scenarios. Then you wouldn't have to worry about the dirty checker or the connect queue. Could result in a nice perf boost as well.

Thoughts?

cc @EisenbergEffect

JeroenVinke:
Sounds great! How would you like to proceed @jdanyow? If you can give me some pointers I can probably create a new PR

jdanyow:
The [`SyntaxInterpreter`](https://github.com/aurelia/templating-binding/blob/master/src/syntax-interpreter.js) is used to interpret binding commands (foo.bind, foo.two-way, foo.delegate, etc) into `BindingExpression` instances. 

I think it would make sense to try replacing the `SyntaxInterpreter`'s `call`, `delegate`, `trigger` and `capture` methods with versions that do not create a binding expression. No need for event bindings in SSR scenarios.

```js
import {SyntaxInterpreter} from 'aurelia-templating-binding';
SyntaxInterpreter.prototype.call = (resources, element, info, existingInstruction, context) => existingInstruction;
SyntaxInterpreter.prototype.delegate = (resources, element, info, existingInstruction, context) => existingInstruction;
...
...
```

To ensure all bindings are "one-time" I was initially thinking we could make a couple more SyntaxInterpreter changes... 
```js
SyntaxInterpreter.prototype.bind = SyntaxInterpreter.prototype['one-time']; 
SyntaxInterpreter.prototype['one-way'] = SyntaxInterpreter.prototype['one-time']; 
```
This would help but wouldn't ensure all bindings are "one-time". Things like [interpolation binding](
https://github.com/aurelia/templating-binding/blob/df1bc30994572054f362388722c66ce8bfe49936/src/binding-language.js#L89-L101) expressions would require other changes. Bindings that use the binding mode binding behaviors (`foo.bind="something & oneWay"`, etc). Might be safer to do something like this:

```js
import {bindingMode} from 'aurelia-binding';

bindingMode.twoWay = bindingMode.oneTime;
bindingMode.oneWay = bindingMode.oneTime;
```

Thoughts? cc @EisenbergEffect

EisenbergEffect:
@jdanyow If you just overwrite oneWay and twoWay with oneTime, that should handle every scenario, correct? If so, let's do that. We should have a single methods on the API that makes this and the event-related changes so that it's easy to call and configure on the server.

jdanyow:
right

EisenbergEffect:
Ok, @JeroenVinke I'm going to close this PR. Can you put together a new PR according to Jeremy's recommendation?

JeroenVinke:
@jdanyow 

```
import {bindingMode} from 'aurelia-binding';

bindingMode.twoWay = bindingMode.oneTime;
bindingMode.oneWay = bindingMode.oneTime;
```

This would cause all binding modes to be `0`, so when this check happens:

https://github.com/aurelia/templating-binding/blob/ebd3dc28b65a9fac15972281afe7078829a9f8b7/src/interpolation-binding-expression.js#L182-L184

it passes, and I believe that causes the DirtyChecker to be enabled down the line.

EisenbergEffect:
@jdanyow Can you provide guidance on this. We're blocked on releasing server-render until we have a suitable solution here. Any thoughts?

jdanyow:
how about:

```diff
-   if (this.mode === bindingMode.oneWay) {
+   if (this.mode !== bindingMode.oneTime) {
      enqueueBindingConnect(this);
    }
```

JeroenVinke:
Thanks, i've made that change but the dirtychecker is still started unfortunately. Now it seems to happen here: https://github.com/aurelia/binding/blob/d5fd3a679cfc17c639330e5fc974c3669fc8221f/src/binding-expression.js#L105

Because `bindingMode.oneWay` and `bindingMode.twoWay` are set equal to `bindingMode.oneTime`, that two-way binding logic is executed. I could probably change the bind() function so it would work, but I wonder if there are more places where we could have similar problems. Any thoughts?

jdanyow:
@JeroenVinke I just pushed a commit that should help there- let me know if anything else comes up.

Issue #589, opened 2017-04-12
Binding to undefined property using `.bind` should not result in `attr="undefined"`


thomas-darling:
**I'm submitting a bug report**

* **Library Version:**
aurelia-binding@1.2.1
aurelia-templating-binding@1.3.0

* **Browser:**
all

**Current behavior:**

Doing this in a template, where `url` is a view model property with the value `undefined`.
```html
<a href.bind="url">
```
Produces this in the DOM:
```html
<a href.bind="url" href="undefined">
```

**Expected/desired behavior:**

* **What is the expected behavior?**

  When the value of the property is `undefined` or `null`, I would not expect the `href` attribute to be present in the DOM.

  In this case, it's a problem, as there's a difference between a link with and without an `href` - one is clickable, the other is not. In this case, I end up with a link to `http://example.com/undefined`, where
I expected the link to be non-clickable.

  Another common case where this is a problem is the `<img>` element, where `src.bind` results in `src="undefined"`, thus triggering an HTTP request to download an image named `undefined`.

jdanyow:
is there a scenario where you'd want a data-bound image or anchor element to be present in the DOM when the url the element is bound to is null or undefined?

thomas-darling:
Yeah, there is, actually :-)

Normally we could work around it with an `if.bind`, but in my particular case that won't work.

We have a `card` custom element, which has a template along the lines of this:
```html
<template>
    <a href.bind="href">
        <slot></slot>
    </a>
</template>
```
Now, this `card` can be used with or without an href, depending on whether we want it to be a link or not. When no href is set, we don't want that `href` attribute on the anchor, as it makes the browser think is is a link, meaning that users can e.g. right-click and open in a new tab - which won't do what they expect.

However, in this case we can't put an `if.bind` on the anchor, as that won't work with the `slot` inside of it.
Only way to make it work, is to manually observe the `href` property and add/remove the the attribute as needed. That workaround does work, but its cumbersome, and I think the proposed behavior is closer to what would be expected - i.e. _if we don't know what value an attribute should have, then it shouldn't be there_. That behavior would also make the very common use cases with e.g. images safe by default - i.e. no more requests for images named `undefined`.

Makes sense?

jdanyow:
Thanks for the additional info. I think I understand your position. I don't think it's semantically correct to use an anchor without an href. You'd at least need to set it to `#` wouldn't you? Otherwise button would be the appropriate element to use.

[...]

PR #616, opened 2017-08-27
fix(ObserverLocator): Handle special attributes for undefined value

bigopon:
fix aurelia/binding#589


jdanyow:
I wonder, instead of adding a new observer, could we change the behavior of `DataAttributeObserver` and `dataAttributeAccessor` such that  `setValue` always removes the attribute when the value is `null` or `undefined`?

In a separate PR, if it makes sense, we could add logic to handle some well-known boolean html attributes like `disabled`. When the model value is falsy the attribute should be removed.

Thoughts?

cc @EisenbergEffect

EisenbergEffect:
That sounds like a good solution to me. @jdanyow Would you be willing to walk @bigopon through how to do that so he can learn more about the binding engine in the process?

jdanyow:
@bigopon here are the methods that would need to be changed:

https://github.com/aurelia/binding/blob/master/src/element-observation.js#L28
https://github.com/aurelia/binding/blob/master/src/element-observation.js#L42

bigopon:
Did you meant this line [Line 191 - observer locator](https://github.com/aurelia/binding/blob/master/src/observer-locator.js#L191)

```js
  // if object is instance of DOM.Element

  return dataAttributeAccessor; // always return this if it's not observable i.e input-able
  // ignore cases
  if (/^\w+:|^data-|^aria-/.test(propertyName)
    || obj instanceof DOM.SVGElement && this.svgAnalyzer.isStandardSvgAttribute(obj.nodeName, propertyName)) {
    return dataAttributeAccessor;
  }
  if (propertyName in obj) {
    return specialDomAttributeAccesor;
  }

```

and then inside `dataAttributeAccessor` it will have `removeAttribute` if value is `null`/`undefined` ? Same in `DataAttributeObserver` at https://github.com/aurelia/binding/blob/master/src/element-observation.js#L42 ?

jdanyow:
yep- line numbers changed since my comment due to PRs... I should have referenced a specific revision.

bigopon:
I had problem with text in composite-observer spec. It tests on `textContent`, and using `setAttribute` if it's not observable will make it do

```js
element.setAttribute('textContent', value); // result in something <div textcontent="value" />
// instead of
element.textContent = value;
```

Can you give advise ? @jdanyow

jdanyow:
That was a poor choice on my part, try using `id` instead. Text content is typically assigned via interpolation bindings.

bigopon:
I've put in all the changes suggested.
@jdanyow

src/observer-locator.js
```diff
@@ -188,10 +188,11 @@ export class ObserverLocator {
...
       }
-      if (/^\w+:|^data-|^aria-/.test(propertyName)
-        || obj instanceof DOM.SVGElement && this.svgAnalyzer.isStandardSvgAttribute(obj.nodeName, propertyName)) {
-        return dataAttributeAccessor;
-      }
+      return dataAttributeAccessor;
+      // if (/^\w+:|^data-|^aria-/.test(propertyName)
+      //   || obj instanceof DOM.SVGElement && this.svgAnalyzer.isStandardSvgAttribute(obj.nodeName, propertyName)) {
+      //   return dataAttributeAccessor;
+      // }
```
jdanyow: Was this intentional? This would make every binding to a DOM element use attribute observation.

bigopon: The edited part belongs to `getAccessor` in observer locator, not `getObserver` so I think it's valid

jdanyow: This would be a breaking change to getObserver. It's not needed for this PR is it?

bigopon: You meant return this back to previous state, and just keep the change in `dataAttributeAccessor` and `DataAttributeObserver` ? Isn't it same with before the fix, that `<a href.bind="prop"></a>` will still go to `propertyAccessor` ?

jdanyow: Right, we'll need to make this more targeted. I think drastically changing the behavior here should wait until Aurelia vNext.

In the ObserverLocator's `createPropertyObserver` and `getAccessor` methods you'll need to add logic that looks for image.src, anchor.href, etc. Here's an [example](https://github.com/aurelia/binding/blob/master/src/observer-locator.js#L116).

Would be nice to remove all usages of `.toLowerCase()` in these conditions. Instead we could use instanceof or just compare with the uppercase string.

The logic in `createPropertyObserver` and `getAccessor` need to continue to match in terms of when element observers vs a property observer is used.

bigopon: What about we roll out previous fix for DOM special property (prop on the proto) only ? We can move `specialDataAttributeAccessor` part to `observer-locator.js` to and change the test to avoid exporting the new accessor ? It will fix the case `a.href='undefined'` and help avoid breaking change ?

```js
if (obj instance of DOM.Element) {
   ...
   if (/^\w+:|^data-|^aria-/.test(propertyName) || obj instanceof DOM.SVGElement && this.svgAnalyzer.isStandardSvgAttribute(obj.nodeName, propertyName)) {
        return dataAttributeAccessor;
   }
   ... the different is here, where we check if property name is in the element proto
   if (propertyName in obj) {
      return specialDomAttributeAccessor;
   }
}
... same after this point
```

jdanyow: I think we can change `dataAttributeAccessor` and `DataAttributeObserver`. Currently setting null/undefined will result in the attribute value becoming the string `'undefined'`/`'null'` which I doubt anyone expects.

bigopon: Ok, so we will have some more checks like this:

```js
  if (/^\w+:|^data-|^aria-/.test(propertyName) || obj instanceof DOM.SVGElement && this.svgAnalyzer.isStandardSvgAttribute(obj.nodeName, propertyName)) {
        return dataAttributeAccessor;
  }
  if (propertyName === 'href' || propertyName === 'src') {
    return dataAttributeAccessor;
  }
```

Is this what you meant ?

jdanyow: yeah

bigopon:
I've adjusted the observer locator, and the tests. Now it only handles `img -> src` and `a -> href`

obedm503:
any info on when this fix will be released?

StrahilKazlachev:
@obedm503 it is.

EisenbergEffect:
It has already shipped, I believe. @jdanyow can correct me if I'm wrong. Is there an issue with binding 1.5.0 where this is not working?

Issue #604, opened 2017-07-13
[Proposal] Adjust `EventManager` and `Listener` classes


bigopon:
**I'm submitting a proposal**

* **Library Version:**
1.2.1

**Current behavior:**
Whenever Listener `bind` is called, it creates a new function `event => this.callSource(event)`, which will be use as `callback` inside `DefaultEventStrategy`.

This `callback` is then assigned to either `delegatedCallbacks` or `capturedCallbacks` properties on the element listening to the event. The disadvantage of this approach is the higher memory allocation for creating lambda in `Listener`, and only one listener allowed per event type per element. While the memory maybe negligible, the latter disadvantage sometimes makes it hard to consume an element. 

For example:

```html
<!-- my-comp.html -->
<template bindable="message" click.delegate="say(message)">
  <!-- ... -->
</template>

<!-- my-comp-consumer.html -->
<template>
  <require from="./my-comp.html" ></require>
  <my-comp message="hello" click.delegate="doExtraWorkWhenClick(someReference)"></my-comp>
  <!-- ... -->
</template>
```

We will either lose `say(message)` or `doExtraWorkWhenClick(someReference)` due to the current way of how event is registered on an element.

# **Proposal**

Lets start with `EventManager` `addEventListener` method:

**CURRENT**
```js
addEventListener(target, targetEvent, callback, delegate)
```

**PROPOSAL**
```js
// Could be named `addListener` to avoid breaking change ?
addEventListener(target, targetEvent, listener /* instance of class listener */, delegate)
```

Changes required in `Listener` class: **NOT breaking (requires test changes)**

```js
  bind(source) {
    if (this.isBound) {
      if (this.source === source) {
        return;
      }
      this.unbind();
    }
    this.isBound = true;
    this.source = source;

    if (this.sourceExpression.bind) {
      this.sourceExpression.bind(this, source, this.lookupFunctions);
    }
    // this._disposeListener = this.eventManager.addEventListener(
    //  this.target,
    //  this.targetEvent,
    //  event => this.callSource(event),
    //  this.delegationStrategy);

    this._disposeListener = this.eventManager.addListener(
      this.target,
      this.targetEvent,
      this,
      this.delegationStrategy
    );
  }
  
  // new extra class method to handle event, leveraging EventTargetInterface
  // This help avoid creating lambda inside each listener (Great win, in my pov for memory)
  handleEvent(event) {
    this.callSource(event);
  }
```

[...]

PR #620, opened 2017-09-12
refactor(EventManager): Support handle event

bigopon:
This PR is aimed to address change in `Listener` proposed in #604 , to use the instance as listener in `bind` instead of creating a new arrow function.

In my local, test 'connects 100 bindings immediately before queueing rest' in `connect-queue.spec.js` breaks as it only flushes around 90 - 91 in queueMicroTask. I'm not sure why.


jdanyow:
checked out this branch. Tests are working for me. This looks good to merge.

Issue #555, opened 2016-12-21
Memory leak


petronellius:
**I'm submitting a bug report**

* **Library Version:**
1.0.9

**Please tell us about your environment:**
* **Operating System:**
Windows 10

* **Node Version:**
4.4.5

* **NPM Version:**
3.9.5

* **JSPM OR Webpack AND Version**
JSPM 0.16.47

* **Browser:**
Chrome 55.0.2883.87

* **Language:**
TypeScript 2.0

**Current behavior:**
Minimal demo: 
[skeleton-typescript - Copy.zip](https://github.com/aurelia/binding/files/666531/skeleton-typescript.-.Copy.zip)

Steps to reproduce:

1. Open Aurelia skeleton in browser and navigate to Memory leak tab
2. Take a Heap snapshot in Chrome Developer Tools Profiler
3. Click Delete button which remove one cell (aurelia custom element).
4. Take a Heap snapshot in Chrome Developer Tools Profiler and choose Comparison view
5. In Comparison view you can see that one cell object was added and none was deleted.

![1](https://cloud.githubusercontent.com/assets/12825401/21392827/06388054-c792-11e6-8b87-67e913bc3517.PNG)

If I remove one of the nested divs or remove value binding on cell component I get expected result - one cell object was added and two were deleted. 

![2](https://cloud.githubusercontent.com/assets/12825401/21392918/620080ee-c792-11e6-93e1-c4e1f6ee889d.PNG)

**Expected/desired behavior:**

* **What is the expected behavior?**
Cell components that are not attached to DOM anymore should be released.

* **What is the motivation / use case for changing the behavior?**

PR #622, opened 2017-09-24
fix(subscriber-collection): use callable to remove subscribers

### LINKED ISSUE #585: memory leak in subscriber-collection.js in case of complex bindings

* **Library Version:**
1.2.1

* **Operating System:**
Windows 10

* **Node Version:**
6.9.1

* **NPM Version:**
4.0.5

JSPM 0.16.46

* **Browser:**
Chrome 57

* **Language:**
ESNext

**Current behavior:**
When using complex binding expressions inside of a repeat, old subscribers do net get cleaned-up on itemsChanged.
 

The example is a list of 100 items that can be filtered by a search box:
https://gist.run/?id=ed14778cdbeccb3888c6ee0c16b8fc71
Set a breakpoint to: https://github.com/aurelia/binding/blob/1d6e826dbbc401ca21ad0deaf7acca498cd71c8b/src/subscriber-collection.js#L49
and enter a search value, e.g. '1' into the search box.
Each time you change the search value the repeat items change and the size of this._callablesRest grows (~200, ~300, ~500, ...).

Outdated subscriptions are not retained by the framework since they may keep a large amout of memory from being freed by the GC.

Allowing to implementing re-usable UI components which require more complex binding expressions, e.g. a data driven list with search and "add new row" capabilities.

bigopon:
fix #585 #555 

possibly fix aurelia/router#456

### DIFF
--- a/src/subscriber-collection.js
+++ b/src/subscriber-collection.js
@@ -43,13 +43,29 @@ function removeSubscriber(context, callable) {
     this._callable2 = null;
     return true;
   }
-  let rest = this._contextsRest;
+  let contextsRest = this._contextsRest;
+  let callablesRest = this._callablesRest;
   let index;
-  if (!rest || !rest.length || (index = rest.indexOf(context)) === -1 || this._callablesRest[index] !== callable) { // eslint-disable-line no-cond-assign
+  if (!contextsRest || !contextsRest.length) {
     return false;
   }
-  rest.splice(index, 1);
-  this._callablesRest.splice(index, 1);
+  if ((index = callablesRest.indexOf(callable)) === -1) {
+    return false;
+  }
+  if (contextsRest[index] !== context) {
+    let len = callablesRest.length;
+    while (len > index) {
+      if (contextsRest[index] === context) {
+        break;
+      }
+      index++;
+    }
+    if (index === len) {
+      return false;
+    }
+  }
+  contextsRest.splice(index, 1);
+  callablesRest.splice(index, 1);
   return true;
 }
 
--- a/src/subscriber-collection.js
+++ b/src/subscriber-collection.js
@@ -43,29 +43,20 @@ function removeSubscriber(context, callable) {
     this._callable2 = null;
     return true;
   }
-  let contextsRest = this._contextsRest;
-  let callablesRest = this._callablesRest;
-  let index;
-  if (!contextsRest || !contextsRest.length) {
+  const callables = this._callablesRest;
+  if (callables === undefined || callables.length === 0) {
     return false;
   }
-  if ((index = callablesRest.indexOf(callable)) === -1) {
+  const contexts = this._contextsRest;
+  let i = 0;
+  while (!(callables[i] === callable && contexts[i] === context) && callables.length > i) {
+    i++;
+  }
+  if (i >= callables.length) {
     return false;
   }
-  if (contextsRest[index] !== context) {
-    let len = callablesRest.length;
-    while (len > index) {
-      if (contextsRest[index] === context) {
-        break;
-      }
-      index++;
-    }
-    if (index === len) {
-      return false;
-    }
-  }
-  contextsRest.splice(index, 1);
-  callablesRest.splice(index, 1);
+  contexts.splice(i, 1);
+  callables.splice(i, 1);
   return true;
 }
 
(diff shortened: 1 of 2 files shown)


jdanyow:
@bigopon could you describe what the issue was and what the fix is? I'm not clear on the scenario.

jdanyow:
disregard my previous comment- I see the issue now. When finding the "rest" index we weren't checking for a match in both contexts and callables.

here's another take- merging some of the conditions... not sure if it's any better or worse.

```js
  let rest = this._contextsRest;
  if (!rest || !rest.length){
    return false;
  }

  let index = 0;
  while ((index = rest.indexOf(context, index)) !== -1) && this._callablesRest[index] !== callable) { }
  if (index === -1) {
    return false;
  }
  rest.splice(index, 1);
  this._callablesRest.splice(index, 1);
  return true;
```

bigopon:
@jdanyow I think it will work in the same way without specifying a start index. We can use the scenario given in issue #585 to check it:

```html
<!-- simplified version of bug report -->

<div repeat.for='item of items'>
  <my-el config.bind='item.prop[$config]'></my-el>
</div>
```

the subscriber list of `$config` will be something like `'sourceContext'` * 100. But the callable will not be the same in the way string is. So `(index = res.indexOf(context, index))` will be equal 1, instead of 0.

jdanyow:
re: edit 2 in https://github.com/aurelia/binding/pull/622#issuecomment-331747634:

ok, let's consider getting rid of indexof altogether. Seems like we could simplify the logic by using while or indexof but not both. Also, since we know context tends to be the same while callable varies more, maybe we could optimize around that heuristic by comparing callable first, followed by context. For example, the && expression in these types of statements would be flipped:  `if (this._context0 === context && this._callable0 === callable) {` becomes  `if (this._callable0 === callable && this._context0 === context) {`.

The indexof + while could be replaced with a single while loop, which compares callable first:

```js
const callables = this._callablesRest;
if (callables === undefined || callables.length === 0) {
  return false;
}
const contexts = this._contextsRest;
let i = 0;
while (callables[i] !== callable || contexts[i] !== context) {
  i++;
}
if (i >= callables.length) {
  return false;
}
contexts.splice(i, 1);
callables.splice(i, 1);
return true;
```

thoughts?

*note: this is code I typed into the comments box for illustration purposes, not tested...*

bigopon:
That looks a lot better !

I tested it, it runs indefinitely because we merge both conditions together. I think it just need a little tweak:

```js
const callables = this._callablesRest;
if (callables === undefined || callables.length === 0) {
  return false;
}
console.log('callables not undefined');
const callableCount = callables.length;
const contexts = this._contextsRest;
let i = 0;
while (callables[i] !== callable && callableCount > i) {
  i++;
}
while (contexts[i] !== context && callableCount > i) {
  i++;
}
if (i === callables.length) {
  return false;
}
contexts.splice(i, 1);
callables.splice(i, 1);
return true;
```

In either way, we can't really avoid a length comparison. The reason is if neither `callable` nor `context` in parameters matches the callable or context at 0 index, it will quit immediately. I feel really hard to explain why it runs indefinitely but it happens if we merge conditions.

jdanyow:
ah the dangers of coding in the comment box :)

```js
const callables = this._callablesRest;
if (callables === undefined || callables.length === 0) {
  return false;
}
const contexts = this._contextsRest;
let i = 0;
while (!(callables[i] === callable && contexts[i] === context)) {
  i++;
}
if (i >= callables.length) {
  return false;
}
contexts.splice(i, 1);
callables.splice(i, 1);
return true;
```

bigopon:
Awesome, also made me realize I put in wrong test 😰 

Doing it this way, however, cause the risk of it running forever if it fail to find a matching pair. should we put a constrain for the length ?

jdanyow:
Ha yeah needs len check

bigopon:
Tested, all good with length, even when I tried to call `removeSubscriber` with invalid params 👍 

One last thing though, `indexOf` seems to have better performance than `while`, should we resolve around it, or `while` works better in your benchmark ?

jdanyow:
How are you comparing? What are you comparing? Which browsers?

To be thorough we should benchmark the different implantations. A jsperf would be fine. I’m good with merging the solution we’ve discussed and attempting to optimize further in another PR. In the past when I’ve tested these types of things there was a wide variation in indexof perf across browsers we support.

bigopon:
I see. All I did was a quick google for [jsperf of for vs indexOf vs while](https://jsperf.com/js-for-loop-vs-array-indexof/10)

Let's do it in future, when things get more consistent

jdanyow:
Sounds good. That jsperf isn't a good fit for our use-case because it's testing an array of integers which is a special case in some browsers, plus we have a two-dimensional comparison, and their "while" test is quite different than ours. I'll open another issue to track the benchmarking.

bigopon:
Great. Code & Test are ready I think. Please have a final check

jdanyow:
looks good- tests passing. Another nice one :+1:

Issue #617, opened 2017-08-27
Binding Changes Property


TonyLugg:
I am submitting a bug report

When I have a view bound to an object and then replace the object with a new object (record), if the properties bound to input boxes are null, they are set to empty string after initial binding.  I have created a sample gist [here](https://gist.run/?id=e4e10b86b5723be0701b1a33903e1577&sha=1ecc4c0e81b801301641957933be1772c73e6a8d)

I have also tested this using routing to load new records, but it gives the same result.  I understand that an input box only understands string and not null, but the initial binding should not alter the bound property unless the user changes something.  This issue affects my ability to determine if something has changed.  Our back end API differentiates null from empty string, because null means the user never entered anything, whereas empty string would indicate there was a value and the user removed it.

bigopon:
A work around is

```html
<input id="txtDepartment" type="text" value.one-way="clientData.department" input.delegate='clientData.department = $event.target.value' />
```

I don't think in your case it is initial binding anymore, you are jumping from one to next. But it could use some improvement, to notify subscriber about binding information, like

```js
callSubscriber(newValue, oldValue, sourceObj, propertyName)
```

So subscribers can decide if it should update. In this case, the model observer see that it's coming from an `input` element, and propertyName is `value`, so no need to update the model with `null` / `undefined` / `empty string` when `oldValue` is either of those. This could be related to type inference @EisenbergEffect @jdanyow

bigopon:
@TonyLugg Another work around for you is:

```js
  btnNext_click() {
    // Following two lines will help cleanup things in the DOM, so your model doesn't get updated back
    this.clientData = null;
    this.originalClientData = null;

    this.getClientData(this.getNextId()).then(clientData => {
      this.clientData = clientData;
      this.originalClientData = JSON.parse(JSON.stringify(clientData));
    });
  }
```

TonyLugg:
@bigopon Thank you for your insight and work arounds.  I have handled it for now in my function that determines changes.  I like your work around to set the objects to null.  I'm wondering if it could be done immediately before setting them to the newly downloaded record object or if there would be timing issues.  I'll test it and report back.

jdanyow:
html input elements coerce all values to empty string. It would be best to make your model property's initial value empty string rather than null. If you need to use null to represent empty input, you could create a value converter whose `toView` replaces null with empty string and whose `fromView` replaces empty string with null.

TonyLugg:
[...]

Issue #443, opened 2016-06-30
databound input changes bound value to string


kuechlerm:
With view and model (typescript):

```
// html
<template><input type="number" value.bind="val"></template>

// ts
class Foo {
  val = 5; // number

  increment() { 
    this.val += 1; 
    console.log(this.val, typeof this.val);
  }
}
```

Now calling twice increment (via button.click) results in
> 6 "number"
> 61 "string"

I expect 
> 6 "number"
> 7 "number"

Using `this.val++` works.
Using one-way binding works.

atsu85:
You could use 2way valueconverter to convert string value from view to number

fopsdev:
or use a custom bindingbheaviour so you are not running into:
http://stackoverflow.com/questions/37882659/force-view-to-be-updated-from-within-a-valueconverter

kuechlerm:
I understand your solution, but this is still confusing and unexpected.
Why is there a difference between `this.val += 1` and `this.val++` ?

fopsdev:
@kuechlerm hmm good question. well i would say that TS implies when using this.val++ that you are operating on a numeric value

atsu85:
@kuechlerm 
> Why is there a difference between this.val += 1 and this.val++ ?

As @fopsdev wrote
> JS implies when using this.val++ that you are operating on a numeric value and therefore treats it like that internally

it is totally JS thing, that has nothing to do with Aurelia. You can verify it out on browser console as well (for example on this site, that doesn't use Aurelia)

EisenbergEffect:
Closing this since the ValueConverter recommendation address the original issue and the following discussion expounds on the JavaScript oddity.

MJLHThomassen:
> Closing this since the ValueConverter recommendation address the original issue and the following discussion expounds on the JavaScript oddity.

And what is this ValueConverter recommendation?

I'm having a similar issue as original poster:

`<input id="nrOfLicensesInput" class="form-control" type="number" value.two-way="nrOfLicenses" min="1" placeholder="1"/>`

Then, whever i change the input box number (either manually or trough the arrows provided by Chrome), in my VM, the nrOfLicenses field is suddenly a string instead of a number. I'm using typescript so i also cant work around it by doing a manual Number.parse(nrOfLicenses) because nrOfLicenses is already a string and I get Typescript erros.

Any suggestions?

EisenbergEffect:
[...]

Issue #442, opened 2016-06-29
Input bindings don't update when changes are reverted within a setter or changeHandler


AdamWillden:
I think these examples will show the issue clearly enough to void more explanation. The interpolated text shows the real value but the inputs don't revert to the underlying value.

getter/setter example: https://gist.run/?id=8c56fe352fd849462d39311ede2c8b5d
computed getter/setter example: https://gist.run/?id=14ce505a70736ca3c76943ac762fe836
changeHandler example: https://gist.run/?id=45e93cac4e74d8ffb702325d56a7192a

I believe the same can be said for textboxes too. I have alarm limits who's ranges are checked (which is related to this). If the limit is at 100, it cannot exceed 100 and 101 is entered it is reverted to a value of 100 but the input itself doesn't reflect this change.

Any ideas? Is this something that can be fixed in the binding system?

AdamWillden:
I implemented a work around by using a binding signaler on the input - I don't consider this an acceptable 'fix' however.

Below is code for the numeric limit input. I only signal when the value has been clamped back to the same old value (`clamped && limit === oldLimit`).

```html
<input value.bind="limitConfig.margin & signal:limitConfig.marginSignal & updateTrigger:'blur'">
```

```js
  limitChanged(limit, oldLimit) {
    if (typeof limit !== 'number') {
      limit = Number(limit);
      this.limit = limit;
    }

    let clamped = false;

    if (this.enabled) {
      if (this.lowerLimitConfig !== null && this.lowerLimitConfig.upperBound > limit) {
        limit = this.lowerLimitConfig.upperBound;
        clamped = true;
      }

      if (this.upperLimitConfig !== null && this.upperLimitConfig.lowerBound < limit) {
        limit = this.upperLimitConfig.lowerBound;
        clamped = true;
      }
    }

    if (clamped) {
      this.limit = limit;
    }

    if (clamped && limit === oldLimit) {
      this.bindingSignaler.signal(this.limitSignal);
    }
  }
```

fopsdev:
A beforeUpdate(newValue,oldValue, cancel) would be nice here
I've opened an issue some while ago with a very similar request:
https://github.com/aurelia/templating-resources/issues/137
The answer is once more: a custom bindingbehaviour :)

[...]

Issue #261, opened 2015-12-15
Value converter and two-way binding - unexpected behavior


StrahilKazlachev:
I have a field `userInput: string` bound to an input. And a clear button that sets the `userInput` to `''`. If a value converter is added I get the following:
* after the view is loaded and no changes are made, the clear button works as expected - `toView` is called with the new value `''`, then `fromView` is called.
* after the view is loaded and the user enters a new value, `userInput` is correctly updated. But now if the clear button is pressed `userInput` is not set to `''`, you need to press it a second time to set the value to `''`. I changed the value converter to just return the value in both `toView` and `fromView`. What I see is that the first time the clear button is pressed `toView` is not called and `fromView` is called with the last value written in the input.

EisenbergEffect:
@jdanyow Can you take a look at this?
@StrahilKazlachev Can you provide a simple reproduction that we can use to see the behavior?

StrahilKazlachev:
I tried, and couldn't, it seems I'm one patch release behind in the project. Updating now.

StrahilKazlachev:
Here is a [repro](https://github.com/StrahilKazlachev/skeleton-navigation) - welcome page. But it looks like I'm doing some things that "confuse" the binding engine. Like `<template if.one-time="!config.readonly">` where `config` is `undefined` and the beginning and I assign it a value in the `bind` callback.

StrahilKazlachev:
@jdanyow It seems I forgot to "mention" anyone, sorry if you've already seen it.

jdanyow:
@StrahilKazlachev I wasn't able to reproduce this issue using the code you provided. What does `jspm inspect --forks` say?

StrahilKazlachev:
@jdanyow I tried now in FF and it worked, but in chrome(47) I still get that behavior.

PR #623, opened 2017-09-26
feat(observable): coercion, typed observable

bigopon:
This PR

  * adds coerce support for decorator `observable`
  * adds ability to create custom `observable` with fluent syntax: `@observable.custom`
  * provide 4 base custom observables: `@observable.date`, `@observable.number`, `@observable.string`, `@observable.boolean`

Usage:

**With normal syntax**

```js

class MyViewModel {
  @observable({ coerce: 'number' }) numProp
  @observable({ coerce: 'boolean' }) boolProp
  @observable({ coerce: val => convertValue(val) }) customCoerceProp
}
```

**Using metadata**

```js

import {metadata} from 'aurelia-metadata';

class MyViewModel {
  @observable
  @Reflect.metadata(metadata.propertyType, Number)
  num
}

var instance = new MyViewModel();
instance.num = '4';
instance.num; // <====== 4
```

TypeScript users can achieve above result (metadata) by simpler code:

```ts
class MyViewModel {
  @observable num: number
}

var instance = new MyViewModel();
instance.num = '4';
instance.num; // <===== 4 , with compile error though
```

Instruction for the compiler to emit decorator metadata [TypeScript decorator doc](https://www.typescriptlang.org/docs/handbook/decorators.html)

##

All coerce type will be resolved to a string, which then is used to get the converter function in `coerceFunctions` export of this module. So, to extend or modify basic implementations:

```js
import {coerceFunctions} from 'aurelia-binding';

// Modify built in
coerceFunctions.string = function(a) {
  return a === null || a === undefined ? '' : a.toString();
}

// Extend
coerceFunctions.point = function(a) {
  return a.split(' ').map(parseFloat).slice(0, 2);
}

// Usage of 'point' coerces defined above:
class MyLine {
  @observable({ coerce: 'point' }) point1
  @observable({ coerce: 'point' }) point2
}
```

For TS users or JS users who want to use metadata, to extend coerce mapping:

```ts

import {
  createTypedObservable
} from 'aurelia-binding';

// use static class method
class Point {
  static coerce(value) {
    return new Point(value);
  }
}
mapCoerceFunction(Point, 'point');

// or just pass a 3rd parameter, fallback to static coerce method when 3rd param omitted:
mapCoerceFunction(Point, 'point', val => new Point(val));

// decorator usage:
// TypeScript
class MyLine {
  @observable point1: Point
  @observable point2: Point
}
// JavaScript
class MyLine {
  @observable
  @Reflect.metatata('design:type', Point)
  point1

  // or like this
  @observable({ coerce: 'point' })
  point2
}
```

##

**With fluent syntax**

```js
class MyLine {
  @observable.number x1
  @observable.number() y1

  @observable.number() x2
  @observable.number y2
}

var line = new MyLine();

line.x1 = '15'
line.x1; // <======= 15
```

To built your own fluent syntax observable:

```js

import {
  coerceFunctions,
  createTypedObservable
} from 'aurelia-binding'

coerceFunctions.point = function(value) {
  return value.split(' ').map(parseFloat);
}
createTypedObservable('point');

// usage:
class MyLine {
  @observable.point point1;
  @observable.point point2;
}
```

### DIFF
--- a/src/coerce-functions.js
+++ b/src/coerce-functions.js
@@ -14,7 +14,7 @@ export const coerceFunctions = {
   boolean(a) {
     return !!a;
   },
-  date(a) {
+  date(val) {
     // Invalid date instances are quite problematic
     // so we need to deal with it properly by default
     if (val === null || val === undefined) {
--- a/src/coerce-functions.js
+++ b/src/coerce-functions.js
@@ -15,7 +15,14 @@ export const coerceFunctions = {
     return !!a;
   },
   date(a) {
-    return new Date(a);
+    // Invalid date instances are quite problematic
+    // so we need to deal with it properly by default
+    if (val === null || val === undefined) {
+      return null;
+    }
+    const d = new Date(val);
+    const t = d.getTime(); // to deal with invalid date
+    return t === t ? d : null;
   }
 };
 
@@ -30,7 +37,7 @@ export const coerceFunctionMap: Map<{new(): any}, string> = new Map([
  * Map a class to a string for typescript property coerce
  * @param type the property class to register
  * @param strType the string that represents class in the lookup
- * @param coerceFunction coerce function to register with @param strType
+ * @param coerceFunction coerce function to register with param strType
  */
 export function mapCoerceFunction(type: {new(): any, coerce?: (val: any) => any}, strType: string, coerceFunction: (val: any) => any) {
   coerceFunction = coerceFunction || type.coerce;
--- a/src/coerce-functions.js
+++ b/src/coerce-functions.js
@@ -1,0 +1,45 @@
+import * as LogManager from 'aurelia-logging';
+
+export const coerceFunctions = {
+  none(a) {
+    return a;
+  },
+  number(a) {
+    const val = Number(a);
+    return !isNaN(val) && isFinite(val) ? val : 0;
+  },
+  string(a) {
+    return '' + a;
+  },
+  boolean(a) {
+    return !!a;
+  },
+  date(a) {
+    return new Date(a);
+  }
+};
+
+export const coerceFunctionMap: Map<{new(): any}, string> = new Map([
+  [Number, 'number'],
+  [String, 'string'],
+  [Boolean, 'boolean'],
+  [Date, 'date']
+]);
+
+/**
+ * Map a class to a string for typescript property coerce
+ * @param type the property class to register
+ * @param strType the string that represents class in the lookup
+ * @param coerceFunction coerce function to register with @param strType
+ */
+export function mapCoerceFunction(type: {new(): any, coerce?: (val: any) => any}, strType: string, coerceFunction: (val: any) => any) {
+  coerceFunction = coerceFunction || type.coerce;
+  if (typeof strType !== 'string' || typeof coerceFunction !== 'function') {
+    LogManager
+      .getLogger('map-coerce-function')
+      .warn(`Bad attempt at mapping coerce function for type: ${type.name} to: ${strType}`);
+    return;
+  }
+  coerceFunctions[strType] = coerceFunction;
+  coerceFunctionMap.set(type, strType);
+}
--- a/src/coerce-functions.js
+++ b/src/coerce-functions.js
@@ -21,8 +21,7 @@ export const coerceFunctions = {
       return null;
     }
     const d = new Date(val);
-    const t = d.getTime(); // to deal with invalid date
-    return t === t ? d : null;
+    return isNaN(d.getTime()) ? null : d;
   }
 };
 
--- a/src/decorator-observable.js
+++ b/src/decorator-observable.js
@@ -35,7 +35,7 @@ export function observable(targetOrConfig: string | Function | ObservableConfig,
       if (coerceFunction === undefined) {
         observableLogger.warn(`Invalid coerce instruction. Should be either one of ${Object.keys(coerceFunctions)} or a function.`);
       }
-    } else if (_usePropType) {
+    } else if (_usePropertyType) {
       propType = metadata.getOwn(metadata.propertyType, target, key);
       if (propType) {
         coerceFunction = coerceFunctions[coerceFunctionMap.get(propType)];
@@ -164,15 +164,15 @@ class     | config           | config
 /**
  * Internal flag to turn on / off auto pickup property type from metadata
  */
-let _usePropType = false;
+let _usePropertyType = false;
 
 /**
  * Toggle the flag for observable to auto pickup property type from metadata
  * The reason is sometimes we may want to use prop type on bindable, but not observable
  * and vice versa
  */
-observable.usePropType = (shouldUsePropType: boolean) => {
-  _usePropType = shouldUsePropType;
+observable.usePropertyType = (shouldUsePropType: boolean) => {
+  _usePropertyType = shouldUsePropType;
 };
 
 /**
--- a/src/decorator-observable.js
+++ b/src/decorator-observable.js
@@ -1,6 +1,56 @@
-export function observable(targetOrConfig: any, key: string, descriptor?: PropertyDescriptor) {
-  function deco(target, key, descriptor, config) { // eslint-disable-line no-shadow
-    // class decorator?
+import { coerceFunctions, coerceFunctionMap } from './coerce-functions';
+import { metadata } from 'aurelia-metadata';
+import * as LogManager from 'aurelia-logging';
+
+/**
+ * @typedef ObservableConfig
+ * @prop {string} name
+ * @prop {string} changeHandler
+ * @prop {string | {(val: any): any}} coerce
+ */
+
+const observableLogger = LogManager.getLogger('aurelia-observable-decorator');
+
+export function observable(targetOrConfig: string | Function | ObservableConfig, key?: string, descriptor?: PropertyDescriptor) {
+  /**
+   * @param target The class decorated
+   * @param key The target class field of the decorator
+   * @param descriptor class field descriptor
+   * @param config user's config
+   */
+  function deco(target: Function, key?: string, descriptor?: PropertyDescriptor & { initializer(): any }, config?: ObservableConfig) { // eslint-disable-line no-shadow
+    // Used to check if we should pickup the type from metadata
+    const userDidDefineCoerce = config !== undefined && config.coerce !== undefined;
+    let propType;
+    let coerceFunction;
+
+    if (userDidDefineCoerce) {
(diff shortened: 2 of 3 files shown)


jdanyow:
Very nice :muscle:- have you had a chance to try this out in data-binding scenarios? I imagine there must be some gotchas when you attempt to use this with two-way binding. What are the limitations that folks should know about? What are some good example use-cases we can put in the docs for this feature?

bigopon:
@jdanyow For two way data binding on text field, it is normal with string, ok-ish with number, and broken for any other type (date, boolean).

I feel like there's a lot of gotchas, but cannot really describe them. I've prepared a gist at https://gist.run/?id=7d59c430a382b3423a13d7f760eb2c00 to demo common cased for typed observable and bindable.

** There is a small modification for date coerce function

```js
date: function(val) {
  // instead of return new Date(val);
  if (val === null || val === undefined) {
    return null;
  }
  const d = new Date(val);
  const t = d.getTime(); // to deal with invalid date
  return t === t ? d : null;
}
```

What do you think ?

EisenbergEffect:
@jdanyow Can you review this?
@bigopon Can you confirm that all syntax is the same for the typed bindables as well?

bigopon:
@jdanyow @EisenbergEffect 
I just discovered a seemingly problematic behavior when using property type with metadata: We are defining the metadata on the prototype, not the constructor. Which will affect serialization scenarios, and probably inheritance too. Example Gist (Check the console, you will see `__metadata__` on the prototype of `App`)
https://gist.run/?id=7d59c430a382b3423a13d7f760eb2c00

Explanation:

When doing:

javascript using `Reflect.metadata`:

```js
class MyClass {
  @Reflect.metadata('design:Type', Number) age
}
```

or 

typescript: emit metadata:

```ts
class MyClass {
  age: number
}
```

We are doing basically this:

```js
Reflect.defineMetadata('design:Type', Number, MyClass.prototype, 'age');
```

nashwaan:
Wow @bigopon . 
This feature will solve and simplify many scenarios.

💪💪💪💪💪

bigopon:
@jdanyow @EisenbergEffect sorry false alarm. Information about coercion for inheritance in both `@observable` and `@bindable` are already carried to derived class (slightly different ways) so it will work fine. Added tests for inheritance for `@observable` in this PR though.

One thing should be discussed is how we want the API for turn on / off the flag to auto pickup type metadata. Two options atm:

```ts
let _usePropertyType = false;
export function usePropertyType(shouldUsePropertyType: boolean) {
  _usePropertyType = shouldUsePropertyType;
}

// OR
observable.usePropertyType = function(shouldUsePropertyType: boolean) {
  _usePropertyType = shouldUsePropertyType;
}
```

Working on `bindable` now and should be able to do some documentation soon.

src/coerce-functions.js
```diff
@@ -15,7 +15,14 @@ export const coerceFunctions = {
...
   date(a) {
-    return new Date(a);
+    // Invalid date instances are quite problematic
+    // so we need to deal with it properly by default
+    if (val === null || val === undefined) {
+      return null;
+    }
+    const d = new Date(val);
+    const t = d.getTime(); // to deal with invalid date
+    return t === t ? d : null;
```
atsu85: is this `t === t` an optimization compared to `!isNaN(t)`, or why do you use this weird syntax, that only makes sense in JavaScript?

bigopon: Yes it is used to specifically deal with NaN value. As `NaN === NaN` is `false`.

What a spec

StrahilKazlachev: But does it perform better than using `isNaN`?

bigopon: Thats interesting question. Normally i would say yes but V8 is smart and I believe isNaN could be faster ( from another v8 engineer tweet i saw different between `o === o` and `Object.is(o, o)` is 2 us

atsu85: @bigopon 
> As `NaN === NaN` is `false`.

Yeah i know, after figuring that out I started using TsLint ;)

My comment+question was because of readability & preformance. IMHO if `t === t` is not preforming (much) better than `!isNaN(t)`, then i'd vote for better readability :)

bigopon:
I see. Cant commit on phone 😁 Will change

src/coerce-functions.js
bigopon: I see. Cant commit on phone 😁 Will change

bigopon: @atsu85 thanks for the suggestion. I just keep doing those things 😄

bigopon:
Demo for this PR + typed bindable is at http://aurelia-typed-observable-bindable.surge.sh/

IE 11 compatible so we can test the binding scenarios for all supported browsers.

bigopon:
@EisenbergEffect This PR is base for aurelia/templating#558, and is waiting for some review from @jdanyow . Beside one thing that is I can no longer break it when setting coerce (and create new object to assign) on both side of two way binding, which is quite unexpected.

If this is good to merge then the other can be merged straight away I believe

EisenbergEffect:
@bigopon I didn't quite follow that last bit. Are you saying that two-way patterns work now as well?

jdanyow:
@bigopon, @EisenbergEffect it would be great to have a few reference use-cases for coerce so we can validate our implementation works for them. I'm worried coerce will cause a bunch of confusion due to the view and view-model values not matching and the odd scenarios that will result, similar to what happens when folks attempt to use a value converter or a property setter to make a number/date control. If you look through the binding issues folks have opened there are a bunch related to folks being surprised that the html input element's value property coerces to string.  Here's a few:
https://github.com/aurelia/binding/issues/442
https://github.com/aurelia/binding/issues/617
https://github.com/aurelia/framework/issues/288
https://github.com/aurelia/binding/issues/261
https://github.com/aurelia/binding/issues/443

bigopon:
@EisenbergEffect Just rechecked it, it is working as expected, In the gist I wasn't declaring `@observable.date` / `@bindable.date` on both side of binding, that's why it is not freezing. So my gotchas above are good. It's just the reference use-cases @jdanyow mentioned.

src/coerce-functions.js
```diff
@@ -0,0 +1,51 @@
+import * as LogManager from 'aurelia-logging';
+
+export const coerceFunctions = {
+  none(a) {
+    return a;
+  },
+  number(a) {
+    const val = Number(a);
+    return !isNaN(val) && isFinite(val) ? val : 0;
```
CasiOo: I would expect either NaN or Infinity to be returned if not a valid number, not 0.

bigopon: `NaN` doesn't work nicely with `===`, to handle it we need to sacrifice perf for more checks. With hundreds of bindings perf hit will be worse. I'm not sure what to do here. `Infinity`, on the other hand, still problematic but could be treated as is.

CasiOo: I see it is a problem with two-way binding. Still, as a developer I would not expect 0 to be returned

src/coerce-functions.js
```diff
@@ -0,0 +1,51 @@
...
+    const val = Number(a);
+    return !isNaN(val) && isFinite(val) ? val : 0;
+  },
+  string(a) {
+    return '' + a;
+  },
+  boolean(a) {
+    return !!a;
+  },
+  date(val) {
```
CasiOo: Since dates are stateful, should there be a check if it's already of type date?
I haven't thought this through, just want to make sure this case has been considered.
```
if (typeof val === 'date') {
  return val;
}
```

src/coerce-functions.js
bigopon: I think it could be application specific where `NaN` / `Infinity` are expected. For that case, we can always define another decorator via `createTypedObservable`:

```js
import {
  coerceFunctions,
  createTypedObservable
} from 'aurelia-binding'

coerceFunctions.$number = function(value) {
  return Number(value);
};
createTypedObservable('$number');
```

// usage
```js
class App {
  @observable.$number() numProp
}
```

src/coerce-functions.js
bigopon: This could be nice. I didn't think of it. Maybe we can reuse the instance ? @jdanyow / @EisenbergEffect

gheoan:
Aurelia does not provide any value converters so why would the coerce functions be provided?
All basic type constructors can be used as functions for type conversion. Why use `@observable({ coerce: 'number' }) prop` instead of `@observable({ coerce: Number }) prop`.
Is there any use case for the fluent syntax other then saving a few keystrokes?
Sorry to delay this PR any longer. I sugest splitting this PR, leaving one PR with only the basic functionality (the coerce attribute accepting a function) so that it can be merged faster.

bigopon:
@gheoan Imo, it is good to have interface used for dependencies declaration. In JS, there is no interface so string is the best option. Plus there are unexpected behavior when you call primitive constructor directly, for example

```js
Date(new Date()) !== new Date(new Date())
```

And how do you override existing / built in implementation if everything is a function ? It is sometime desired to do so. For example, in this PR implementation, we have identical behavior between:

```js
class ABC {
  @observable({ coerce: 'string' }) value 
}
```

```js
class ABC {
  @observable({ coerce: String }) value
}
```

because they both dont care `null` / `undefined`. It will just be converted as is. Now in someone's app, it's better to have type `string`, but `null`/ `undefined` to an empty string. Enforcing function would 
require the converter function to be imported everywhere, which is what I don't believe we want to do, like following:


```js
import { RealStringConverter } from './util'

class ABC {
  @observable({ coerce: RealStringConverter }) value
}
```

While with string:


```js
// main.js
import { coerceFunctions } from 'aurelia-binding'
coerceFunctions.string = val => val === null || val === undefined ? '' : val.toString();

// app.js
class ABC {
  @observable({ coerce: 'string' }) value
}
```

Does the above answer your question ?

---

I do want to see the PR get merged, but like @jdanyow stated, it would be nice if we could have real world usage to validate the implementation. It works fine in tests but the experience could vary in app.

bigopon:
@EisenbergEffect @jdanyow  Maybe this could be enabled via plugin form, so people can try it out and before we include it in the box ?

jdanyow:
agree- plugin would be better for something like this.  I'm worried that it will be hard to support given all the issues mentioned in https://github.com/aurelia/binding/pull/623#issuecomment-350594177

rek72:
Thanks @bigopon for the plugin.  My only question is that if we use this, is there much concern over future aurelia binding releases?  Are you all rewriting this or just extending the binding from aurelia?

Issue #353, opened 2016-03-21
Value converters should support triggering updates


thomas-darling:
Ok, so here we go...
### Background

I understand the concept of binding behaviors, and think they are a really great idea for things like throttling binding updates - for that stuff they make a lot of sense. However, the more I work with this, the more I find myself wrapping a value converter in a binding behavior, just to make sure it updates when the user changes site settings such as locale or currency.

The `TBindingBehavior` in the Aurelia i18n library is a perfect example of this:
https://github.com/aurelia/i18n/blob/master/src/t.js

Now, we are a large, global site and unfortunately we found the Aurelia i18n library to be quite insufficient for our needs - that's a topic for another discussion that I would love to have with you guys, but the point is that we are working on a custom library for this, and in doing that, we need to implement a number of custom value converters to handle formatting of things like currencies, dates, times, numbers, percentages, pluralization, lists, prepositions, word variants, and lots of other crazy stuff.

The problem is, every single one of those converters need to be wrapped in a corresponding binding behavior, as it needs to update whenever the locale or currency is changed. Actually, I can't think of very many uses for value converters where such a requirement would not be present - after all, their fundamental purpose is to convert values to/from their view representation, and that will almost always be a localized string.

I feel like this workaround is an abuse of the binding behavior concept, as what we're doing here feels much more like a value conversion than like a behavioral change. And if every converter has to be wrapped in a behavior anyway, then why even have the distinction? It just blurs the line between the two concepts, and ultimately makes it a bit of a mess.
### Proposal

The way I see it, binding behaviors should be used only for things like throttling binding updates or setting things like `oneWay` or `oneTime` - uses where they actually modify the fundamental behavior of the _binding_. Meanwhile, value converters should have a way to signal to the underlying binding that it needs to update - it does not need to have full access to the actual binding instance, it just needs some way to signal to the binding that it needs to update... or maybe even simpler, just a property that we can specify, which lists the signals that should trigger an update (you can think of it as just stating dependencies on signals, just like the constructor states dependencies on services, etc.)

Example:

``` typescript
export class CurrencyValueConverter
{
    // When the framework instantiates the binding converter, it should look for this property,
    // and if present, it should subscribe the binding to the specified signals.
    signals = ['currency-changed', 'locale-changed'];
    ...
}
```

This would make this whole thing so much nicer and more intuitive to work with, and I think it would make the distinction between the two concepts much more clear.

EisenbergEffect:
[...]

PR #626, opened 2017-10-02
feat(ValueConverter): enable signal

bigopon:
This PR:

  * Addresses #353 Value converters should support triggering updates
  * Has potential integration with `BindingSignaler` in `templating-resources`, where signaling in there will also trigger here, treating signal universal
  * Still needs tests (WIP)
  * Usage:

  ```js
  export class LocaleValueConverter {
    signals: string | string[] = 'locale-changed'
  }

  // usage
  import { ConverterSignaler } from 'aurelia-binding';
  ConverterSignaler.signal('some-thing');
  ```

  * [Demo](https://gist.run/?id=e99a86c05beea939b9c900bf533f59d4)
  * [Updated Demo](https://gist.run/?id=2f3f613fc5a1ca100e9f881e9feb0dce)

@jdanyow @EisenbergEffect @thomas-darling @davismj

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -157,8 +157,6 @@ export class ValueConverter extends Expression {
     if (signals === undefined) {
       return;
     }
-    // support both input type 'signal' & ['signal-1', 'signal-2']
-    signals = Array.isArray(signals) ? signals : [signals];
     i = signals.length;
     while (i--) {
       connectBindingToSignal(binding, signals[i]);
--- a/src/ast.js
+++ b/src/ast.js
@@ -1,5 +1,41 @@
 import {Unparser} from './unparser';
 import {getContextFor} from './scope';
+import {sourceContext} from './connectable-binding';
+
+/**
+ * @typedef IBinding
+ * @prop {{()}} bind
+ * @prop {{()}} unbind
+ * @prop {{()}} call
+ */
+
+/**@type {{[x: string]: Set<IBinding>}} */
+const converterBindingSignals = {};
+
+function callBinding(binding: IBinding) {
+  binding.call(sourceContext);
+}
+
+export const ConverterSignaler = {
+  addBinding(name: string, binding: IBinding): void {
+    const bindings = converterBindingSignals[name] || (converterBindingSignals[name] = new Set());
+    if (!bindings.has(binding)) {
+      bindings.add(binding);
+    }
+  },
+  removeBinding(name: string, binding: IBinding): void {
+    const bindings = converterBindingSignals[name];
+    if (bindings !== undefined) {
+      bindings.delete(binding);
+    }
+  },
+  signal(name: string): void {
+    let bindings = converterBindingSignals[name];
+    if (bindings !== undefined) {
+      bindings.forEach(callBinding);
+    }
+  }
+};
 
 export class Expression {
   constructor() {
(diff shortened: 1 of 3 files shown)


jdanyow:
It would be nice if we could unify this API with the existing binding signaler API. I think there's an opportunity here to tie signaling in at a lower level, in the "connect" phase and refactor the existing signaling logic to leverage the existing binding machinery related to property observation.

Today, the ValueConverter expression class looks like this:

```js
export class ValueConverter extends Expression {
  ...
  ... other methods ...
  ...
  connect(binding, scope) {
    let expressions = this.allArgs;
    let i = expressions.length;
    while (i--) {
      expressions[i].connect(binding, scope);
    }
  }
}
```

I think we could change it to something like this:

```js
import { connectBindingToSignal } from './signals';

export class ValueConverter extends Expression {
  ...
  ...
  ...
  connect(binding, scope) {
    const expressions = this.allArgs;
    let i = expressions.length;
    while (i--) {
      expressions[i].connect(binding, scope);
    }
    // new stuff:
    const signals = binding.lookupFunctions.valueConverters(this.name).signals;
    if (signals === undefined) {
      return;
    }
    i = signals.length;
    while (i--) {
      connectBindingToSignal(binding, signals[i]);
    }
  }
}
```

*aurelia-binding - signals.js (new module)*

```js
const signals = Object.create(null);

export function connectBindingToSignal(binding, name) {
  if (!signals.hasOwnProperty(name)) {
    signals[name] = 0;
  }
  binding.observeProperty(signals, name);
}

export function signalBindings(name) {
  if (signals.hasOwnProperty(name)) {
    signals[name]++;
  }
}
```

Then our [BindingSignaler](https://github.com/aurelia/templating-resources/blob/master/src/binding-signaler.js) class in templating-binding would reuse the new `signalBindings` function. We should probably mark it obsolete and inform users to switch over to `signalBindings`.

The signal binding behavior would become much simpler by reusing then new `connectBindingToSignal` function.

What do you think?

bigopon:
You meant we determine signals based on `BindingBehavior` / `ValueConverter` -> `signals` properties ?

bigopon:
@jdanyow Things looks a lot neater now. very cool 👍 💯 

  * templating-resources also ready.
  * What do you think about the tests ?

src/ast.js
```diff
@@ -148,6 +149,20 @@ export class ValueConverter extends Expression {
...
+    let converter = binding.lookupFunctions.valueConverters(this.name);
+    if (!converter) {
+      throw new Error(`No ValueConverter named "${this.name}" was found!`);
+    }
+    let signals = converter.signals;
+    if (signals === undefined) {
+      return;
+    }
+    // support both input type 'signal' & ['signal-1', 'signal-2']
+    signals = Array.isArray(signals) ? signals : [signals];
```
jdanyow: This is a really hot path. Let's remove this line entirely to ensure the connect overhead remains minimal.

I don't think supporting string & array is necessary, just more to document & maintain. If we were to do that here's a good spot to perform the normalization, but again, I would prefer to keep the API surface small and strict:

https://github.com/aurelia/binding/blob/b6565eb116c212188feec296ae4782c3c1d95c89/src/value-converter-resource.js#L15-L17

bigopon: @jdanyow All done as suggested. Learnt to be more careful with `connect`. About the API, 99% of usage will be a single signal, I do think we need to support normalisation, but I'll let you make the call after this got merged. What do you think ?

EisenbergEffect:
@bigopon Would you be willing to PR some documentation for this feature in the value converter doc?

bigopon:
Sure i enjoy doing it, and discussing with Jeremy too 😁

PR #627, opened 2017-10-03
refactor(Element Observation): use handleEvent instead of bind

bigopon:
This PR replaced

  * `checked-observer.js`, `select-observer.js`: `this.synchronizeValue.bind(this, false)` with `handleEvent() -> this.synchronizeValue()`
  * `ValueAttributeObserver`: this.notify.bind(this, false)` with `handleEvent() -> this.notify()`

I'm don't see bound parameter `false` used anywhere in `synchronizeValue` functions so I think it's safe to remove. Tests remain the same.

### DIFF
--- a/src/checked-observer.js
+++ b/src/checked-observer.js
@@ -111,9 +111,13 @@ export class CheckedObserver {
     this.callSubscribers(newValue, oldValue);
   }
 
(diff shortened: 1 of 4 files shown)


jdanyow:
nice!

PR #630, opened 2017-10-08
refactor(bindingMode): replace oneWay with toView

### LINKED ISSUE #629: cleanup: replace oneWay with toView
We should standardize the internals on `toView` now that it has landed. Here's one instance, probably others in `templating-binding`, `templating`, `templating-resources`, etc.

https://github.com/aurelia/binding/blob/b6565eb116c212188feec296ae4782c3c1d95c89/src/binding-expression.js#L104

tdeschryver:
Closes #629.


doc/article/en-US/binding-basics.md
```diff
@@ -25,10 +25,10 @@ Aurelia supports binding HTML and SVG attributes to JavaScript expressions. Attr
 * `attribute`:  an HTML or SVG attribute name.
 * `command`: one of Aurelia's attribute binding commands:
   * `one-time`: flows data one direction: from the view-model to the view, **once**.
-  * `one-way` / `to-view`: flows data one direction: from the view-model to the view.
+  * `to-view`: flows data one direction: from the view-model to the view.
```
jdanyow: I think we should leave reference to `one-way` so folks will know that `one-way` and `to-view` are the same thing.  `to-view` should definitely be the preferred command.

src/aurelia-binding.d.ts
```diff
@@ -134,17 +134,17 @@ export declare interface CollectionObserver {
...
   oneTime = 0,
   /**
    * Updates the binding target (target) property when the binding source (source) changes.
    * This type of binding is appropriate if the element being bound is implicitly read-only.
-   * If there is no need to monitor the changes of the target property, using the OneWay
+   * If there is no need to monitor the changes of the target property, using the to-view
    * binding mode avoids the overhead of the two-way binding mode.
    */
-  oneWay = 1,
+  toView = 1,
```
jdanyow: The addition of `toView` here is good. We need to keep `oneWay` here for now, it's not fully deprecated.

src/binding-mode.js
```diff
@@ -1,7 +1,6 @@
 export const bindingMode = {
   oneTime: 0,
-  oneWay: 1,
```
jdanyow: Need to keep `oneWay` here for now, it's not fully deprecated.

doc/article/en-US/binding-basics.md
tdeschryver: OK, I wasn't sure if it should be fully removed or not.

doc/article/en-US/binding-basics.md
```diff
@@ -25,10 +25,10 @@ Aurelia supports binding HTML and SVG attributes to JavaScript expressions. Attr
 * `attribute`:  an HTML or SVG attribute name.
 * `command`: one of Aurelia's attribute binding commands:
   * `one-time`: flows data one direction: from the view-model to the view, **once**.
-  * `one-way` / `to-view`: flows data one direction: from the view-model to the view.
+  * `to-view` / `one-way`: flows data one direction: from the view-model to the view.
```
atsu85: perhaps it should me made clear, that
> to-view should definitely be the preferred command

(as @jdanyow said) - in the other words, `one-way` is deprecated (but can still be used for backwards compatibility reasons), but `to-view` is the preferred command?

jdanyow:
thanks @tdeschryver!

PR #631, opened 2017-10-09
doc(signals): doc for value converter signals

bigopon:
This PR:

  * Fixes code listing for `capture` doc in delegate vs capture
  * Adds doc for value converter signals with example


jdanyow:
nice example!

Issue #635, opened 2017-10-15
performance: remove toLowerCase


jdanyow:
We have a couple modules where we use `toLowerCase` unnecessarily. We don't need the extra string allocation, especially in these hot paths. Would be great to get rid of these.

* https://github.com/aurelia/binding/blob/master/src/event-manager.js
* https://github.com/aurelia/binding/blob/master/src/observer-locator.js

PR #637, opened 2017-10-19
performance: remove toLowerCase #635

rodrigoheffner:
Removed variable allocations as described on issue #635.

### DIFF
--- a/src/event-manager.js
+++ b/src/event-manager.js
@@ -221,11 +221,11 @@ export class EventManager {
     let properties = config.properties;
     let propertyName;
 
-    this.elementHandlerLookup[config.tagName.toLowerCase()] = {};
+    this.elementHandlerLookup[config.tagName] = {};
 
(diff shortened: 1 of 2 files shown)


jdanyow:
@rodrigoheffner thanks for picking this up! I'd like to get rid of the `.toLowerCase()` calls where possible. There are several places where we do things like `element.tagName.toLowerCase() === 'select'` that could change to `element.tagName === 'SELECT'` because we know the DOM will uppercase the `tagName` property:

> On HTML elements in DOM trees flagged as HTML documents, `tagName` returns the element name in the uppercase form.
https://developer.mozilla.org/en-US/docs/Web/API/Element/tagName

rodrigoheffner:
Makes sense, @jdanyow - I'll update my changes with that in mind, thanks!

harryrf:
Is this issue still open?

PR #641, opened 2017-10-28
Add support for non-ASCII characters in identifiers

### LINKED ISSUE #640: Binding to Non-ASCII identifier throws Lexer Error: Unexpected character
## Context

When you talk to non-english customers about their non-english domain, having a common [ubiquitous language](http://www.jamesshore.com/Agile-Book/ubiquitous_language.html) gets invaluable real quick. The need to translate back and forth between terms that are actually used in the business and far-fetched or even wrong translations will easily frustrate developers.

Furthermore, having a non-english ubiquitous language [comes with its own quirks](https://www.webfactory.de/blog/ubiquitous-language-in-a-non-english-domain): Developers have to agree upon where to use which language, and how they will be combined in identifiers: Is it `GetItem()` or `GetEintrag()` or even `EintragAbrufen()`? 

If your team decides on non-english identifiers, it will make a big difference being able to use Umlauts, Accents, other diacritics or whatever your language has going for you, and not having to fake it using ASCII characters.

The JavaScript language allows a huge set of unicode characters to be used for identifiers: [This blog post](https://mathiasbynens.be/notes/javascript-identifiers-es6) elaborates on what characters may be used and how to generate regular expressions for them. 

## Issue with current implementation

For some developers, it might be desirable to make use of non-ASCII characters in their models or components and bind to them. The current implementation will fail with the following exception:

> Error: Lexer Error: Unexpected character [ü] at column 8 in expression [test.erfüllt ? 'toggle on green ' : 'toggle off ']
    at Scanner.error (vendor.js?v=4RXa-WsgCDo46VzGSv7zYts0C3cRr-5bjPLxuAyWOgM:8557)
    at Scanner.scanToken (vendor.js?v=4RXa-WsgCDo46VzGSv7zYts0C3cRr-5bjPLxuAyWOgM:8389)
    at Lexer.lex (vendor.js?v=4RXa-WsgCDo46VzGSv7zYts0C3cRr-5bjPLxuAyWOgM:8305)
    at new ParserImplementation (vendor.js?v=4RXa-WsgCDo46VzGSv7zYts0C3cRr-5bjPLxuAyWOgM:8711)

## Proposal

The aforementioned blog post shows how to generate code that matches any valid identifier. Unfortunately, it needs to query the web for this task, and the resulting expressions are rather big. Hence, a generic solution for the scanner seems impractical. 

I have thrown together a solution that allows to globally add some characters to the identifier filter, leaving it untouched otherwise. It has no performance impact if the user doesn't call it, and should only have very little impact when few characters are added.

frankebersoll:
Adds a static method `addIdentifierCharacters(string)` that can be used to modify how `isIdentifierStart/Part` work by adding a range of non-ASCII characters.

## Example

```typescript
// boot.ts
import { addIdentifierCharacters } from 'aurelia-binding';

export async function configure(aurelia: Aurelia) {
    addIdentifierCharacters("äöüÄÖÜß");
    (...)
}
```

This would allow binding expressions to use identifiers with any of those characters.

Fixes #640.

### DIFF
--- a/src/lexer.js
+++ b/src/lexer.js
@@ -292,6 +292,24 @@ export class Scanner {
   }
 }
 
+export function addIdentifierCharacters(chars) {
+  const codes = new Set();
+  for (var i = 0; i < chars.length; i++) {
+    codes.add(chars.codePointAt(i));
+  }
+
+  const _isIdentifierStart = isIdentifierStart;
+  const _isIdentifierPart = isIdentifierPart;
+
+  isIdentifierStart = function (code) {
+    return _isIdentifierStart(code) || codes.has(code);
+  }
+
+  isIdentifierPart = function (code) {
+    return _isIdentifierPart(code) || codes.has(code);
+  }
+}
+
 const OPERATORS = {
   'undefined': 1,
   'null': 1,
@@ -380,14 +398,14 @@ function isWhitespace(code) {
   return (code >= $TAB && code <= $SPACE) || (code === $NBSP);
 }
 
-function isIdentifierStart(code) {
+let isIdentifierStart = function(code) {
   return ($a <= code && code <= $z)
       || ($A <= code && code <= $Z)
       || (code === $_)
       || (code === $$);
 }
 
-function isIdentifierPart(code) {
+let isIdentifierPart = function(code) {
   return ($a <= code && code <= $z)
       || ($A <= code && code <= $Z)
       || ($0 <= code && code <= $9)
--- a/doc/article/en-US/binding-basics.md
+++ b/doc/article/en-US/binding-basics.md
@@ -163,3 +163,40 @@ The binding system makes several properties available for binding in your templa
 * `$last` - In a repeat template, is true if the item is the last item in the array.
 * `$even` - In a repeat template, is true if the item has an even numbered index.
 * `$odd` - In a repeat template, is true if the item has an odd numbered index.
+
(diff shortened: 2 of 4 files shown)


frankebersoll:
Maybe we need another name. I'm not happy with the current one.

frankebersoll:
What needs to be added/changed to make this a thing?

src/lexer.js
```diff
@@ -292,6 +292,24 @@ export class Scanner {
...
+export function addIdentifierCharacters(chars) {
+  const codes = new Set();
+  for (var i = 0; i < chars.length; i++) {
+    codes.add(chars.codePointAt(i));
+  }
+
+  const _isIdentifierStart = isIdentifierStart;
+  const _isIdentifierPart = isIdentifierPart;
+
+  isIdentifierStart = function (code) {
```
jdanyow: `function isIdentiferStart...` should change to `let isIdentiferStart = ...` to signal that it's value could change. Same for `isIdentifierPart`.

frankebersoll: Fixed

fkleuver:
@EisenbergEffect Correct. All 1,350 characters from the [Latin](https://en.wikipedia.org/wiki/Latin_script_in_Unicode#Table_of_characters) script will be supported.

EisenbergEffect:
Closing in favor of #689

PR #643, opened 2017-10-29
Fix signalable-value-converters demo

timfish:
http://aurelia.io/docs/binding/value-converters#signalable-value-converters

It might be a better idea to fix some of the styling issues with the main docs css.


jdanyow:
much appreciated :+1:

PR #646, opened 2017-11-07
perf(Lexer): operators array -> map

bigopon:
Micro benchmark at https://gist.run/?id=26d099040a363b23729807ea6f7bff43

Results:

- on Chrome: map is about 50 times faster
- on Firefox: map is about 100 times faster

Tests remain the same

@jdanyow I think it's fast enough to remove the todo [here](https://github.com/bigopon/binding/blob/63dfcb44244dddc54e3af82e49f504b131a5b709/src/lexer.js#L169-L170)

### DIFF
--- a/src/lexer.js
+++ b/src/lexer.js
@@ -127,7 +127,7 @@ export class Scanner {
 
   scanOperator(start, text) {
     assert(this.peek === text.charCodeAt(0));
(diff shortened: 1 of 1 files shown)


jdanyow:
ohhh good find :+1: Nice work!

Issue #645, opened 2017-11-06
Feature: Support spread operator in view


davismj:
**I'm submitting a feature request**

**Current behavior:**
> Error: Parser Error: Unexpected token . at column 19 in [computePathString(...points)]

**Expected/desired behavior:**
Spreads the points array, as if you had passed `computePathString(points[0], points[1], points[2])`, which also observes these variables.

* **What is the motivation / use case for changing the behavior?**
To observe the items in the array when computing a value.

Alexander-Taran:
https://github.com/aurelia/binding/pull/647#issuecomment-350692094
>For what its worth, I was mostly trying to get a discussion going to see if this was a good idea in the first place. I think the discussion here has mostly proved that it isn't.

@davismj ?

Issue #309, opened 2016-02-05
valueConverter doesn't observe array-valued object properties


davismj:
I'm filtering a list of items of various colors based on an array of selected colors:
```
repeat.for="item of items | filter: { property: 'color', value: selectedColors }
```
But `toView` is not triggered when I add or remove colors to the `selectedColors` array.

davismj:
Specifically, I was adding items to the array using 
```
<input type="checkbox" checked.bind="selectedColors" value="red" />Red
```

davismj:
However, it does work as expected if I do this:
```
this.selectedColors = this.selectedColors.concat(["red"]);
```
Which I guess makes sense, as it is `Object.observe`ing.

davismj:
Different use case, also doesn't work.
```
repeat.for="item of items | filter: selectedColors
```

z-brooks:
Tested it in a new skeleton with the same result. But if you add `selectedColors.length` to the object 
```
filter: { ..., update: selectedColors.length }
```
it updates on push/etc.  Or you could use `signal` to trigger the update... :worried:

davismj:
I think one solution that might solve it would be to enhance the property observation strategy when the property is an array. So, [right here](https://github.com/aurelia/binding/blob/master/src/observer-locator.js#L158), we would add the following

```
} else if (obj[propertyName] instanceof Array) {
    return new SetOrModifyArrayObserver(this.dirtyChecker, obj, propertyName);
}
```
Which would combine the behaviors of the `SetterObserver`, setting the subscriptions on the object, notifying on referential changes, and the `ModifyArrayObserver`, notifying on record add/removes. 

What do you think, @jdanyow? Is this a viable strategy?

davismj:
@z-brooks Thanks for checking, dude! I'm always finding out that my issues are due to my own faults in code, so it's relieving to have the review!

jdanyow:
hmm this should work- I had enhanced the repeat not too long ago to add support for this very feature.  There are tests for this scenario here: https://github.com/aurelia/templating-resources/blob/master/test/repeat-integration.spec.js#L388

@davismj could you put together a plunker that reproduces this?  If you feel like looking into the issue, the repeat should be hitting this piece of code:

https://github.com/aurelia/templating-resources/blob/master/src/repeat.js#L180

What it's doing is "unwrapping" the expression so that it observes `items` for mutations instead of the array returned from the value converter (which wouldn't be worth observing)

davismj:
[...]

PR #647, opened 2017-11-11
feat(AST): Spread operator

bigopon:
From feature request: #645 

### Supported syntax:

```html
<div>
  <!-- spread object literal -->
  <el prop.bind="{ ...{ one: 1 } }"></el>

  <!-- spread object accessed by key -->
  <el prop.bind="{ ...someObject }"></el>

  <!-- spread literal array -->
  <el prop.bind="[...[5,6,7, someValue]]"></el>

  <!-- spread array value accessed by key -->
  <el prop.bind="[...someArray]"></el>

  <!-- Call function with spread literal array -->
  <el prop.bind="getValue(...[5,6,7, someValue])"></el>

  <!-- Call function with spread accessed by key -->
  <el prop.bind="getValue(...someArguments)"></el>
</div>
```

### Observation strategy

* Only the accessed name is observed. Not every property of the object or array is observed automatically.

* [Demo](https://gist.run/?id=7722c6f3e9b302985e4ccf321bbd38f8)

@jdanyow What other tests should be in 😄 ?
@davismj @jdanyow Did I miss any usage of spread operator ?

### DIFF
--- a/src/ast.js
+++ b/src/ast.js
@@ -598,12 +598,23 @@ export class LiteralArray extends Expression {
 
   evaluate(scope, lookupFunctions) {
     let elements = this.elements;
+    let length = elements.length;
     let result = [];
-
-    for (let i = 0, length = elements.length; i < length; ++i) {
-      result[i] = elements[i].evaluate(scope, lookupFunctions);
+  
+    for (let i = 0, eli = 0; eli < length; ++eli) {
+      let element = elements[eli];
+      if (element.isSpread === true) {
+        let value = element.evaluate(scope, lookupFunctions);
+        let jj = value.length;
+        for (let j = 0; j < jj; ++j) {
+          result[i + j] = value[j];
+        }
+        i += jj;
+      } else {
+        result[i++] = element.evaluate(scope, lookupFunctions);
+      }
     }
-
+  
     return result;
   }
 
@@ -631,11 +642,19 @@ export class LiteralObject extends Expression {
     let instance = {};
     let keys = this.keys;
     let values = this.values;
-
+  
     for (let i = 0, length = keys.length; i < length; ++i) {
-      instance[keys[i]] = values[i].evaluate(scope, lookupFunctions);
+      let key = keys[i];
+      if (key.isSpread === true) {
+        let value = key.evaluate(scope, lookupFunctions);
+        for (let prop in value) {
+          instance[prop] = value[prop];
+        }
+      } else {
+        instance[key] = values[i].evaluate(scope, lookupFunctions);
+      }
     }
-
+  
     return instance;
   }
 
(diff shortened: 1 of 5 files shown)


jdanyow:
@bigopon nice work. As far as the implementation goes, it looks pretty good to me.

I'm worried about adding something like this without being able to support common scenarios.  I'm afraid it will cause a lot of confusion when simple things like `foo.bind="{ ...myArray }"` doesn't update when the array mutates or when `foo.bind="[ ...myObj ]` doesn't update when a property is added.

Today we have a almost entirely clean distinction between property observation and collection mutation observation.  Adding something like this is either going to fall short of expectations or force us to mix property observation and collection observation with a healthy dose of dirty-checking to cover all the scenarios.

bigopon:
Re: distinction between prop / collection: what was your solution / or there seemed to be no good clear solutions so we stop mixing them ?

About spread: if we cannot overcome the language limitation and have to add dirty checking, then that's more of educating users than a concern, I think. But having restricted usage should be healthy for Aurelia. So we close this ?

jdanyow:
What I meant regarding the prop/collection observation was in aurelia collection observation never happens automatically... you have to opt-in by using the repeat or the binding engine directly. This has been really nice for performance.

Well... almost never... we have a pain-in-the-neck feature where we support `${myArray}` printing the contents of the array and observing the array for changes.  This is the one exception where we look at the value of a property, see that it's an array, and add on array observation. The code for this is [here](https://github.com/aurelia/templating-binding/blob/master/src/interpolation-binding-expression.js), it's gotten a lot simpler since we added the `@connectable` decorator. This is from the very early days, when binding was rapidly evolving and we were trying to accommodate tons of feature requests. https://github.com/aurelia/framework/issues/77 

Anyways, it seems to me that doing rest/spread "right" would require us to start automatically observing collections.  Maybe thats ok if we do it only in the context of the spread AST node's `connect` method. Not sure... I don't want to cause confusion where the binding system behaves unpredictably when it comes to collection observation.  In the past we've resisted requests to do things like this. Discussion here: https://github.com/aurelia/binding/issues/309#issuecomment-183105155

Thoughts?

cc @EisenbergEffect

bigopon:
I think only observing in the context of Spread is good. But I'm not sure if it will make anything unstable. We can still create a plugin to have automatic collection observation if needed anyway. Does a plugin sound good ? @jdanyow @EisenbergEffect

bigopon:
The previous was not clear, I meant we can create a plugin and do automatic array observation there, so if anyone wants to trade performance for dev experience, they can use the plugin. In that case, just need to export some Ast utility functions.

bigopon:
@EisenbergEffect Would be nice if you could input some thoughts here too

davismj:
For what its worth, I was mostly trying to get a discussion going to see if this was a good idea in the first place. I think the discussion here has mostly proved that it isn't.

PR #658, opened 2018-01-12
refactor(ObserverLocator): avoid anonymous fn

bigopon:
- add `EventSubscriber` class
- change internal of `CheckedObserver`, `ValueAttributeObserver`, `SelectValueObserver`
- tests remain same

@jdanyow @EisenbergEffect

### DIFF
--- a/src/observer-locator.js
+++ b/src/observer-locator.js
@@ -31,12 +31,20 @@ import {SVGAnalyzer} from './svg';
 export class ObserverLocator {
   static inject = [TaskQueue, EventManager, DirtyChecker, SVGAnalyzer, Parser];
 
+  /**
+   * @param {TaskQueue} taskQueue
+   * @param {EventManager} eventManager
+   * @param {DirtyChecker} dirtyChecker
+   * @param {SVGAnalyzer} svgAnalyzer
+   * @param {Parser} parser
+   */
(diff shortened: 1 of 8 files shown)


bigopon:
@jdanyow before we expose the typings for observers in #656 in the next release, can you review this so we have a chance to correct typings to avoid a breaking change later ? This PR is created to make element value observation more effective

src/observer-locator.js
```diff
@@ -81,7 +89,8 @@ export class ObserverLocator {
     return value;
   }
 
-  addAdapter(adapter: ObjectObservationAdapter) {
+  /**@param {ObjectObservationAdapter} adapter */
+  addAdapter(adapter) {
```
EisenbergEffect: Is there some reason that you remove the typing?

bigopon: probably to make the file javascript only, while preserving typing via jsDoc as there are no type annotation in the rest of the file. If can be either way for TypeScript, which is what I thought we use for generating typings. But since we are crafting to definition manually, this should be ok.

bigopon: If this PR is good to merge then I can push another commit to revert the typing here, also to adjust the typing in `CheckedObserver`, `SelectObserver` introduced in #656

bigopon: @EisenbergEffect I already fixed the typing issue in those observer constructors, so dis regard my previous comment.

PR #655, opened 2018-01-07
feat(typescript): add ICollectionObserverSplice definition

jonathaneckman:
Closes #652

PR #659, opened 2018-01-24
fix(typescript) ICollectionObserverSplice support for Map

jonathaneckman:
PR #655 only accounted for the splice returned when observing an array. This fix includes Map observer splices. 

Confirmed looking at the existing typings that no other collection type is currently supported.


jonathaneckman:
Standby. Just saw that the typings were out of date. The code also supports Set. Might as well get this in as well.

jonathaneckman:
Disregard my Set changes. I am unable to test them because of a PAL issue [tracked here](https://github.com/aurelia/pal/issues/19). I will submit Set support once I get through that issue.

jonathaneckman:
Closing in favor of #661

PR #669, opened 2018-02-28
fix(doc): css-attribute without interpolation explained

Alexander-Taran:
explaines https://github.com/aurelia/templating-binding/issues/100


bigopon:
Maybe it's worth noting that `css` was introduced to support IE11, as IE11 doesn't like style attribute with interpolation syntax.

bigopon:
Same thing about `<require />` element, anything that seems to off standard keyword should be clarified with documentation.

bigopon:
Same thing about `<require />`  instead of `<import />` element, anything that seems to off standard keyword should be clarified with documentation.

PR #671, opened 2018-03-12
perf(EventManager): enable dispose() pattern for addEventListener

bigopon:
This PR enable `dispose()` on `EventManager.addEventListener`, which will help reduce memory usage for aurelia app, regardless `delegate`, `capture` or `trigger` commands used, as it will not return an anonymous function, but a disposable object instead. Breaking change: no, only return `Disposable` when the fifth parameter is truthy, can be made `true` only to have stricter API.

@jdanyow @EisenbergEffect

### DIFF
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -115,8 +115,10 @@ export declare class EventManager {
    * @param targetEvent Name of event to subscribe.
    * @param callback Event listener callback.
    * @param delegate True to use event delegation mechanism.
+   * @param diposable True to return a disposable object with dispose() method instead of a function
    * @returns function wich removes event listener.
    */
+  addEventListener(target: Element, targetEvent: string, callback: (event: Event) => any, delegate: delegationStrategy, disposable: true): Disposable;
   addEventListener(target: Element, targetEvent: string, callback: (event: Event) => any, delegate: delegationStrategy): () => void;
 }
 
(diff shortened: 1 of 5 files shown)


bigopon:
this together with #658 should benefit Aurelia applications memory well if released together.

EisenbergEffect:
@bigopon Were you able to profile or otherwise test this and #658 together to ensure this is actually making an improvement in memory or perf?

bigopon:
I'm having some issues following the instruction in `README.md` of this benchmark https://github.com/krausest/js-framework-benchmark . Not sure if I can provide perf benchmark result.

Beside perf reason, i think it is reasonable to have this update, as after this, all (should be) subscriptions will implement `dispose()` method, which could help for vNext cleanup ? Around 25 ~ 30 mangling friendly added.

src/aurelia-binding.d.ts
```diff
@@ -115,8 +115,10 @@ export declare class EventManager {
    * @param targetEvent Name of event to subscribe.
    * @param callback Event listener callback.
    * @param delegate True to use event delegation mechanism.
+   * @param diposable True to return a disposable object with dispose() method instead of a function
    * @returns function wich removes event listener.
    */
+  addEventListener(target: Element, targetEvent: string, callback: (event: Event) => any, delegate: delegationStrategy, disposable: true): Disposable;
```
EisenbergEffect: Shouldn't the default for `disposable` be `false` in order to maintain backwards compatibility.

bigopon: It's only the typing, which tells we only return Disposable when and only when the last parameter is `true`, not just truthy. This is to strict the API and avoid breaking changes, as if someone used this API (doubt this  ) their 5th parameter needs to be `true` to have their code potentially break.

EisenbergEffect: Oh! Gotcha. I looked at it in isolation and didn't realize it was providing a signature specific for that code path. Cool!

PR #677, opened 2018-04-07
fix(SelectValueObserver): observe characterData mutation

### LINKED ISSUE #521: Select bound to a sub-property does not update properly when changing the parent property
- **Library Version:**
  1.0.6
- **Browser:**
  Chrome, Edge

I have a Select element which is bound to a sub-property: 
`<select value.bind="selectedCar.selectedModel">`
- The initially displayed value (as set in the constructor) is correct.
- If I change the value of `selectedCar` - the select always displays the first option, no matter what the value of `selectedCar.selectedModel` is. A string interpolation next to the select displays the correct value.
- If I change only the value of the subproperty : `selectedCar.selectedModel` - the select displays the correct option.

See this gist that reproduces the problem: https://gist.run/?id=9537cfd05a9c8c3666be8646343248b8

### LINKED ISSUE #614: Setting select options and value at the same time
Gist: https://gist.run/?id=285933c1cb0e2ff4b7fe67879db946ae

Would this be considered a bug, or is it impossible to accomplish in this way?

It seems like a simple case. Is there an alternative?

---
Edit: I managed to work around the issue with something like this:

```ts
const matchGame = { /* the data from an API call */ };

// Assign this data first, so that the select options can be bound before the value.
this.matchGame.opponent1 = matchGame.opponent1;
this.matchGame.opponent2 = matchGame.opponent2;

this.matchGame = matchGame;
```

bigopon:
fixes #521
fixes #614
fixes aurelia/framework#717
fixes aurelia/templating-binding#124

----

recheck https://github.com/aurelia/i18n/issues/147

The existing mutation observer config works in situations where the number of `<option/>` elements changes, but fails where only the text content of those element change. Probably because of this, we didn't get a many reports from folks who use hierarchical select bindings.

all fixes demo at https://fix-select.bigopon.surge.sh/

@jdanyow @EisenbergEffect

### DIFF
--- a/src/select-value-observer.js
+++ b/src/select-value-observer.js
@@ -155,7 +155,7 @@ export class SelectValueObserver {
       this.synchronizeOptions();
       this.synchronizeValue();
     });
-    this.domObserver.observe(this.element, { childList: true, subtree: true });
+    this.domObserver.observe(this.element, { childList: true, subtree: true, characterData: true });
   }
 
   unbind() {
--- a/test/select-value-observer.spec.js
+++ b/test/select-value-observer.spec.js
@@ -1,13 +1,8 @@
 import './setup';
-import {DOM} from 'aurelia-pal';
-import {bindingMode} from '../src/binding-mode';
-import {
-  createElement,
-  checkDelay,
-  createObserverLocator,
-  getBinding
-} from './shared';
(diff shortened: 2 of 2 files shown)


Alexander-Taran:
@bigopon - superhero!!!

Issue #30, opened 2015-02-13
ability to add logging info when binding to non-existent properties


devmondo:
the idea is that sometimes we bind to a none existent property, and the logger should tell us that  this property does not exist.

but one of the great things about Aurelia is that sometimes you want to bind to dynamic properties, so making this an optional flag in logging verbosity level or something would make it better.

jdanyow:
related to https://github.com/aurelia/binding/issues/23

jdanyow:
maybe it would also be useful to log when dirty-checking is used in a browser with O.o support.

jdanyow:
this came up in the gitter again today:  https://gitter.im/Aurelia/Discuss?at=55140f63c25916c13d35183d

andreister:
I tried to look into this but got stuck at adding logging to `property-observation.js`

I changed the code there to inject the logger (similar to how it's done in other places) 
```
System.register(["aurelia-logging"], function (_export) {
   var LogManager, _logger, ...
   ...
  return {
    setters: [
      function(_aureliaLogging) {
        LogManager = _aureliaLogging;
      }
    ],
    ...
```
and updated `jspm_packages\github\aurelia\binding@0.4.0\.jspm.json` to include the new dependency, but this didn't work - until I also mentioned the new dependency in my app's `config.js`

```
System.config({
  "map": {
     ...
    "github:aurelia/binding@0.4.0": {
      "aurelia-logging": "github:aurelia/logging@0.2.6",
       ...
    },
   ...
```

why is that? Am I missing anything, or it's really necessary to update the main `config.js`?

jdanyow:
Hey- thanks for looking a this, you're on the right track.  What you're running into is aurelia/binding doesn't have a dependency on aurelia/logging yet.  You'd need to `jspm install aurelia-logging' (which would add the entries in config.js).  config.js is typically maintained by jspm, usually you don't need to hand edit it.

The logging code might go higher up, in the `ObserverLocater` class, and should be configurable.  Reason is certain observer classes are used in certain scenarios, depending on the browser's support for O.o, etc.

andreister:
Thanks for the explanations! Didn't realize `jspm install` is the way to change `config.js`

I'm quite new to all this so will need some more guidance. Can you look at my commit https://github.com/andreister/binding/commit/f616ec88494f22bf82e9a6b42b1e7fffc6a26f28 and see if it's close? 

[...]

PR #679, opened 2018-04-24
refactor(parser): merge lexer with parser & optimize parser

fkleuver:
This is a rewrite of the parser that replaces all string-based token comparisons with bitwise operators to speed things up, and also combines the lexer into the parser.

Parsing binary expressions is flattened into a single recursive function that takes operator precedence into account (drastically reduces unnecessary nested calls).

Functionality should be identical afaik. Error checking/reporting might not be - still need to check various error scenarios to ensure the strictness is the same. I wanted to gather some feedback first before I spend a bunch of time on that though :)

Performance improvement of around 200% depending on the type of expressions parsed. Simple benchmark available [here](https://fkleuver.github.io/binding/).

Overall file size slightly reduced (70 bytes less unminified, I guess a little more minified because I added some comments)

@jdanyow @EisenbergEffect Is this viable you think?


EisenbergEffect:
Wow! Very cool! @jdanyow and @bigopon Can you take a look at this?

bigopon:
I like this, especially binary expressions parsing, it's clever (though I'd say old parser is easier to digest), and tests look nice. Checked out and tested on one of my projects. 1 or 2 places to pick for some more gains. It's @jdanyow @EisenbergEffect to decide what to do with `Lexer` and `Token`

fkleuver:
@bigopon Thanks for taking a look!

> 1 or 2 places to pick for some more gains

Which ones?

Alexander-Taran:
awesome "refactoring"

fkleuver:
Updated number scanning now, for normal integers it will calculate them instead of parse which is about 15x faster. For floats there's no difference. Also removed the offset thing from `error()`. As far as performance goes that's pretty much all the low-hanging fruits now afaik.

fkleuver:
My takeaway from the feedback so far is that it's a nice idea but the readability could be improved, so with the latest commit I cleaned things up a bit.

- Wrapped `input.charCodeAt` in `currentChar` getter (was `peek` in the Lexer)
- Wrapped `index++` in `nextChar()` function (was `advance()` in Lexer)

And a few more misc things where I reuse functions instead of inlining things. 

The inlining didn't give nearly as big a performance boost as the switch from string tokens to bitwise tokens. Runtime of the mini benchmark went from ~116ms to ~124ms (compared to ~360ms without the changes) so I'd argue a few % performance loss is probably worth the gain in readability/maintainability.

Regarding splitting the parser/lexer, I kind of had to throw everything on one big pile first so I could see the whole picture. Main point was to parse while scanning (instead of scanning everything and then parsing) as this saves some memory and array operations.
The classes don't really need to be separate. If you guys prefer it, I could probably split them up again without without any significant performance loss.  That's up to you guys now.

fkleuver:
Side note, the runtime could probably also be sped up a little bit by using the numeric tokens instead of strings in the `Binary.evaluate` (potentially with a function lookup table instead of a switch). Downside is during debugging you see numbers instead of strings like '===' if you inspect the AST. Opinion on this @bigopon @jdanyow @EisenbergEffect ?

EisenbergEffect:
I'm ok with having it as one  class or splitting it. No real hard opinion from me. It it's easier to maintain together, then I think it's fine to keep it as one. I'd like to get another pass from @bigopon on this and maybe some feedback from @jdanyow as well, to make sure there aren't any missed edge cases that for some reason weren't covered under existing tests.

I'm pretty excited about this :)

jdanyow:
This looks awesome- nice work :muscle: :rocket:

My only feedback is around the benchmark- I'd like to see the test cases skewed more towards typical binding use cases, just to confirm we don't have a regression there (I doubt we will).  Bindings using AccessScope and AccessMember expressions outnumber other types of expressions by a large factor in typical apps.  Could you update the benchmark link to at least 4 times as many simple expressions like `fooA`, `fooB`, `fooC`, ... , `fooZ` and `foo.barA`, `foo.barB`, etc?

src/parser.js
```diff
@@ -84,52 +90,52 @@ export class ParserImplementation {
...
 
-      while (this.optional(':')) {
-        // TODO(kasperl): Is this really supposed to be expressions?
+      while (this.optional(T_Colon)) {
+        // todo(kasperl): Is this really supposed to be expressions?
         args.push(this.parseExpression());
       }
 
-      result = new ValueConverter(result, name, args, [result].concat(args));
+      result = new ValueConverter(result, name, args);
```
jdanyow: nice.

src/parser.js
```diff
@@ -7,50 +6,57 @@ import {
...
 } from './ast';
 
-let EOF = new Token(-1, null);
-
 export class Parser {
+  cache;
   constructor() {
-    this.cache = {};
-    this.lexer = new Lexer();
+    this.cache = Object.create(null);
```
jdanyow: :+1:

jdanyow:
@fkleuver what are your thoughts on https://github.com/aurelia/binding/pull/641? Is there a better way of handling that in the new Parser architecture? Could we handle it out of the box while keeping backwards compat and not adding too much bloat?

EisenbergEffect:
How big is the lookup table?

fkleuver:
@EisenbergEffect Approximately 10kb for [identifier start](https://github.com/mathiasbynens/unicode-10.0.0/blob/master/Binary_Property/ID_Start/code-points.js) and [identifier part](https://github.com/mathiasbynens/unicode-10.0.0/blob/master/Binary_Property/ID_Continue/code-points.js) (compressed)

fkleuver:
Sure. Could add something like a `ParserOptions` class with a `ScanUnicodeIdentifier` flag/method that gets called if `ScanToken` falls through to the default case.

It doesn't necessarily have to be a standalone plugin btw. As long as the lookup is in a separate file, the build script could output an additional `aurelia-binding-unicode` file that includes it. I've seen @bigopon do some really neat stuff composing a bundle of only the stuff you need. If that works for templating, it should work for binding too :)

fkleuver:
Hm, took a look at it and having an alternative `dist/../aurelia-binding-unicode.js` probably causes more problems than it solves due to the existing hard references in the tooling etc.

I added an [initial idea](https://github.com/fkleuver/binding/commit/033948b7327aed60d4d46d39346a9cfad8e8da44) to do this via a simple overridable extension point. Performance impact for the default implementation is negligible as it only does one extra call on the last character of an identifier.

Users could provide their own implementation and register it as `IdentifierValidator` themselves, or a plugin could do this. That's probably the best of both worlds.

I'm not 100% sure if this is the best name and/or form of encapsulation for the 2 methods. I was thinking it could be better to have a more generic `...Options/Extensions` class injected into the parser, so the signature doesn't need to be changed if extra options need to be added in the future.

Thinking here of things like auto object creation (#205), debug logging (#30), globally adding a value converter (don't recall the issue).

What do you guys think? Have a single `...Options` kind of class injected, or just extend with additional constructor arguments if extra stuff is needed in the future?

jdanyow:
For sure- out of the scope of this PR, just wanted your thoughts on it. Thanks!

fkleuver:
Alright, I removed that last commit.

To tie up any remaining loose ends, there are still 3 `//todo`'s from the old parser that I wasn't sure what to do with.

1. [// todo(kasperl): Is this really supposed to be expressions?](https://github.com/fkleuver/binding/blob/new-parser/src/parser.js#L96)
Can't think of any expression that wouldn't work as parameters to behaviors/converter. Maybe cover the various possibilities with tests and then remove the comment?

2. [// todo(kasperl): Check that this is an identifier. Are keywords okay?](https://github.com/fkleuver/binding/blob/new-parser/src/parser.js#L184)
According to the spec [they are not](https://tc39.github.io/ecma262/#prod-ReservedWord), but in practise every browser seems to be fine with identifiers like `undefined`, `null`, `true`, etc. Excluding those would be a breaking change so I'd say check that it's an [identifier, literal or indexed property](https://tc39.github.io/ecma262/#prod-PropertyName) or a keyword. That should cover all working scenarios I think?

3. [// todo(kasperl): Stricter checking. Only allow identifiers and strings as keys. Maybe also keywords?](https://github.com/fkleuver/binding/blob/new-parser/src/parser.js#L302)
Same as 2.

fkleuver:
Hm, writing a few more tests I noticed that keyed properties on object literals don't work (neither in new nor in current parser). I was trying to get the tests to confirm accurate error messages. 

This: `{['foo']: 'bar'}` (which is valid syntax)

Produces `Missing expected token : at column 1`

I'm considering to fix this with an `AccessKeyed` which is the correct thing to do there from as far as I understand. Any objections to that?

jdanyow:
last two commits look good.  I think the two kasperl todos were holdovers from the original angular dart source for the lexer/parser/ast.

bigopon:
For extension, we can apply the same strategy like Unparser. But will be opt in instead of opt out

huochunpeng:
Damn hardcore! Just want to provide feedback that it all worked in my app.

fkleuver:
Nice, thanks!

stsje:
Will the performance improvement be noticeable in business applications?

jdanyow:
@fkleuver I was getting ready to merge this- noticed a small issue with one of the patterns used in the parser tests that have try-catch logic- I think these changes need to be made to each try-catch style test:

```diff
  describe('does not parse LiteralObject with computed property', () => {
    const expressions = [
      '{ []: "foo" }',
      '{ [42]: "foo" }',
      '{ ["foo"]: "bar" }',
      '{ [foo]: "bar" }'
    ];

    for (const expr of expressions) {z
      it(expr, () => {
+       let error = null;
        try {
          parser.parse(expr);
-         pass = true;
        } catch (e) {
-         expect(e.message).toContain('Unexpected token [');
+         error = e;
        }
+       expect(error).not.toBeNull();
+       expect(error.message).toContain('Unexpected token [');
      });
    }
  });
```

otherwise if the parse doesn't throw as expected, the test will still pass.

fkleuver:
@jdanyow Glad you caught that, apparently there were a couple duplicate tests in there as well and two little issues with the invalid assignment tests. All should be sorted now.

jdanyow:
nice work @fkleuver!

Alexander-Taran:
awesome !!!!11

PR #681, opened 2018-05-04
fix(EventManager): remove unnecessary dereference

bigopon:


Alexander-Taran:
Can you please comment on why it is unnecessary?

jdanyow:
possibly unnecessary but have been erring on the side of caution to avoid memory leaks / aid in GC

bigopon:
The class is a disposable object, which could be de-referenced in a single assignment in its consumer. I think it's common practice, hence this PR

bigopon:
Yes. If its proven to be a source of leaks, we can  add this back. I dont know if anyone else is using this new API since the 5th argument needs to be exactly true. And please note that it needs to be disposed exactly the same way with the old subscription: dispose and dereference the subscription, as the old api returns a function that holds reference to all objects used.

PR #688, opened 2018-05-16
fix(typings): accept Set<any> for collectionObserver

### LINKED ISSUE #687: collectionObserver typings do not accept Set
The Aurelia documentation states that collection observers work for Array , Map , and Set types.

However, the signature of `collectionObserver` is currently specified as
```ts
collectionObserver(collection: Array<any> | Map<any, any>): CollectionObserver;
```
This currently prevents the invocation of with a `Set<any>`.

Mobe91:
fixes #687

### DIFF
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -779,7 +779,7 @@ export declare class BindingEngine {
   /**
    * Gets an observer for collection mutation.
    */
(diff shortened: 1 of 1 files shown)


Alexander-Taran:
@fkleuver can you look @ this one?

fkleuver:
Thanks for adjusting it @Mobe91 

@EisenbergEffect I believe this can be merged without issues

PR #689, opened 2018-05-17
Non-ASCII identifiers, typeof, in, instanceof, templates

### LINKED ISSUE #500: Parser/AST: string interpolation support
@natrim commented on [Fri Sep 09 2016](https://github.com/aurelia/templating-binding/issues/97)

- **Library Version:**
  1.0.0
- **Language:**
  TypeScript

**Current behavior:**
template literals don't work with ternary operator

```
<template bindable="what">
  <div>
    ${what ? `WHAT: ${what}!` : 'what?'}
  </div>
</template>
```

**Expected behavior:**
i expected it to work like

```
<template bindable="what">
  <div>
    ${what ? 'WHAT: '+what+'!' : 'what?'}
  </div>
</template>
```

### LINKED ISSUE #608: Passing $parent to value converter and adding a binding behavior offends the parser.

* **Library Version:**
1.2.0

* **Operating System:**
Windows 10

* **Node Version:**
8.1.0

* **NPM Version:**
5.0.3

JSPM 0.16.53

* **Browser:**
all
* **Language:**
all

**Current behavior:**
When passing `$parent` to a value converter and then using a binding behavior I get an `Error: Unexpected character &.` 

Sample:
https://gist.run/?id=4e78f4b92109035f346caf2ca6153408

Strangely enough, if I wrap it in object notation like so: `{ vm: $parent }` and pass that, it works just fine. This only seems to affect `$parent` and not `$this`. I've not tested any of the other contextual properties.

Expression will parse without error, and behave like the workaround.

### LINKED ISSUE #640: Binding to Non-ASCII identifier throws Lexer Error: Unexpected character

### LINKED ISSUE #685: TypeError: this.sourceExpression.connect is not a function

* **Library Version:**
2.0.0

* **Operating System:**
Windows 10

* **Node Version:**
8.11.1

* **NPM Version:**
5.6.0

* **Browser:**
Chrome 66

* **Language:**
all

**Current behavior:**
When there is an error in binding expression in Aurelia template, like:
`<input value.bind="dto..firstName" />`

Stack trace has no information that would help figure where the error originates from. For example:

> Uncaught TypeError: this.sourceExpression.connect is not a function at ChildInterpolationBinding.connect (MarkerClusterGroup.Refresh.js:110) at enqueueBindingConnect (MarkerClusterGroup.Refresh.js:110) at ChildInterpolationBinding.bind (MarkerClusterGroup.Refresh.js:110) at View.bind (MarkerClusterGroup.Refresh.js:110) at If._show (MarkerClusterGroup.Refresh.js:110) at If._update (MarkerClusterGroup.Refresh.js:110) at If.conditionChanged (MarkerClusterGroup.Refresh.js:110) at BehaviorPropertyObserver.selfSubscriber (MarkerClusterGroup.Refresh.js:110) at BehaviorPropertyObserver.call (MarkerClusterGroup.Refresh.js:110) at BehaviorPropertyObserver.setValue (MarkerClusterGroup.Refresh.js:110) at If.descriptor.set [as condition] (MarkerClusterGroup.Refresh.js:110) at Object.setValue (MarkerClusterGroup.Refresh.js:110) at Binding.updateTarget (MarkerClusterGroup.Refresh.js:110) at Binding.call (MarkerClusterGroup.Refresh.js:110) at BehaviorPropertyObserver.callSubscribers (MarkerClusterGroup.Refresh.js:110) at BehaviorPropertyObserver.call (MarkerClusterGroup.Refresh.js:110)

Example stack trace is for template fetched from server, therefore all source-code line references point to the same, random line of code (`MarkerClusterGroup.Refresh.js:110`) . For bundled template, stack trace is similar, but with source-code line numbers being correct. However, both stack traces do not help developer figure where the error originates from.

Show meaningful message to help developer determine what the problem is.  At least: 

> Invalid binding expression: "dto..firstName"

fkleuver:
This is a fairly large refactoring (again), mostly to merge methods with only a single call to them and overall reduce the call stack depth. Slightly better performance, simpler to debug, and hopefully, easier to read :)

- Adds support for most common non-ASCII characters and allow adding others (fixes #640)
- Adds support for `typeof`, `instanceof`, `in`, `void`
- Adds support for template literals + tagged templates (fixes #500)
- Allow $parent as an argument to variadic expressions (Fixes #608)
- Throw during parsing on invalid member expression starting with `.` (Fixes #685)
- Improved test coverage

The nodejs benchmark script should give more realistic results, using high-resolution timestamps and much lower iterations to reduce the amount of optimization and branch prediction interfering.

Should now be more optimized for initial load (and actually performs a bit worse than my previous version on very high iterations - but again, these are not really realistic)

@jdanyow  @EisenbergEffect  @bigopon


bigopon:
Very nice. 🐛 at https://github.com/aurelia/binding/blob/master/src/parser.js#L786 gone and added support for `instanceof`, `typeof` and `in`, ❤️❤️ but I couldn't find the tests for them. Maybe we need some tests for those operators as well ?

fkleuver:
Right, I forgot about those throughout all the re-rewriting and rewiring :P
I do still need to fix that now; they're not implemented in the AST. Could easily be added in Binary/Unary.
Was thinking about directly assigning the appropriate operator functions to the AST during parsing, so the switch statements don't need to be re-evaluated on every `evaluate`. I dropped that ball as I didn't want to make this PR too big in one go. Anyway, what do you guys think about that? Then I'll just do that right away with adding `typeof`, `instanceof` and `in`. Or just the current way.

fkleuver:
Yep, new commit incoming later. Just tying up some additional loose ends.

fkleuver:
Let's try again shall we :)

Was thinking about just adding all operator keywords since it's only 20-30 extra bytes on the file per operator (maybe less). The ones that make any sense (and are possible) to implement, are `void`, `typeof`, `instanceof` and `in` so I added those. Maybe `void` is overkill in hindsight. Anyone uses it?

**Changes**

This was the last big refactoring, promise :) 
It's now pretty close to what I think is the most lightweight and performant it can be (that is, going any further will turn it into a maintenance nightmare)
There are only 2 recursive parsing methods left: `parseLeftHandSideExpression` (which includes primary, unary, accessor- and call expressions) and `parseExpression` which goes down into ternary -> binary -> leftHandSide.

The context mask that's passed around plays the main role in reducing the recursion and call depth. Downside with bitmasks is they make perfect sense when you're the one writing them, but deciphering someone else's work is quite a bit harder. If any of these need more clarification just let me know :)

@bigopon @jdanyow  @EisenbergEffect

bigopon:
Not easy for me about all this, trying.

About new operators, I think all those are not particularly popular / useful in template expression, but having them makes Aurelia template more powerful, so I think it's worth adding. 

> Maybe void is overkill in hindsight. Anyone uses it?

same with about, plus you can use this to signal Aurelia to observe something, plus passing an `undefined` value at the same time in a single expression, could be handy.

fkleuver:
@bigopon Fair enough, thanks for reviewing.

Small tip for figuring out the flow:
- `gulp build`
- Go to `bench/bench.js` line 180 and change `run(10)` to `run(1)`
- Go to `dist/commonjs/aurelia-binding.js` and set a breakpoint at line 2369 (first line of `ParserImplementation.prototype.parseChain = ...`)
- Start VS Code debugger

Should be easier than either using karma or simply reading the code :)

fkleuver:
@bigopon Template literals and tagged templates now fully implemented (pretty much according to spec)

src/ast.js
```diff
@@ -589,6 +618,52 @@ export class LiteralString extends Expression {
...
+    const results = new Array(this.length);
+    for (let i = 0; i < this.length; i++) {
+      results[i] = this.expressions[i].evaluate(scope, lookupFunctions);
+    }
+    if (this.tagged) {
+      const func = this.func.evaluate(scope, lookupFunctions);
+      if (typeof func !== 'function') {
+        throw new Error(`${this.func} is not a function`);
+      }
+      return func.call(null, this.cooked, ...results);
```
bigopon: it should preserve execution context, instead of `null`. Test:

```js
class Test {
 a(template, exps) {
  console.log(this, template, exps);
 }
}

new Test().a`5${4}3` // this should log: Test instance, ['5', '3', raw: [...]], 4
```

src/parser.js
```diff
@@ -40,554 +37,391 @@ export class ParserImplementation {
...
         continue;
       }
-
-      this.lastIndex = this.startIndex;
       this.startIndex = this.index;
-
-      // handle identifiers and numbers.
-      if (isIdentifierStart(this.currentChar)) {
-        return this.scanIdentifier();
+      if (this.currentChar === /*$*/0x24 || (this.currentChar >= /*a*/0x61 && this.currentChar <= /*z*/0x7A)) {
```
bigopon: Does this mean `_` is not considered an identifier ?

fkleuver: It is. As explained in the comment right below that, that `if` statement (and the one further up for whitespace) is purely a performance tweak for the most common characters. Comparisons are several times faster than an array indexer. The logic itself is redundant.

You could comment them out and reduce the `while` loop in `nextToken()` code to just this, and it would still work:
```js
while (this.index < this.length) {
  this.startIndex = this.index;
  if ((this.currentToken = CharScanners[this.currentChar](this)) !== null) { // a null token means the character must be skipped
    return;
  }
}
```

src/ast.js
fkleuver: Good catch. I fixed this in my latest commit

bigopon:
Very nice. About tagged template literal, it should be consistent with the rest: undefined resilient. That is, when its to display a value, if the func specified is not a function, ignore it. And throw in a call expression like event handler. This way folks would not get hard time using it in their templates. `mustEvaluate` is the param name in call member and scope.

fkleuver:
I do think there should be some form of logging in those cases, as this also leads to those dreaded "my element does not render and it's not showing any errors" if I'm not mistaken. But that discussion belongs elsewhere though. I believe with my latest commit it should be consistent with the rest :)

fkleuver:
Actually, just thinking about it for two more seconds I think this might be mistaken. You see, `Call` expressions are not "endpoints" so, like you say, event handlers and such will throw when they receive a `null` function.
`LiteralTemplate` *is* an endpoint: it only consumes the function itself, so there is nothing else that will throw if the `LiteralTemplate` returns null. It's the consumer of the tag function. Isn't just seeing a non-rendered template more of a pain than seeing an error that tells you to fix the tag function?

bigopon:
Tested an app against latest commit with template literal on normal expression and listener expression, working / throwing as expected 👍 About the size, it's more lines but the actual size after minification should be small, since everything in new parser is defined as local variables, thus is friendly for that purpose. LGTM @EisenbergEffect @jdanyow

fkleuver:
@bigopon Thanks :)

FYI, I double checked the minified size. According to https://javascript-minifier.com/ the AMD build went from 93.232kb to 97.087kb minified (gained 3.855kb).
2.635kb of this comes from the `LiteralTemplate` implementation, 0.686kb from the `addIdentifierPart`/`Start` extension point, 0.310kb from the compressed Unicode char codes, a small bit from the added operators and another few hundred bytes from the inlining of char codes and removing local variables in parsing functions (the latter being the usual performance vs. size tradeoff).

EisenbergEffect:
@fkleuver How does this size compare to the original parser before we removed the Lexer?

fkleuver:
If I replace the code in `parser.js`, `unparser.js`, `ast.js` and add the lexer from 1.7.1 that gives a minified size of 94.252kb for the AMD output. So that's a 2.835kb gain compared to 1.7.1.

Some possibilities for reducing the size:
- Merge `PrefixNot` and `PrefixUnary` (and just call it `Unary`?). ~0.3kb
- Shorten the property names on the `ParserImplementation` to 1-3 letters. ~1kb.
- Dismantle `ParserImplementation` (making it non-exportable) into loose functions. I'm experimenting with that [here](https://github.com/fkleuver/aurelia-binding-experiment/blob/master/src/parser.ts) ~1kb
- Remove the extension point for `addidentifierPart` / `Start`. ~0.7kb (I'm saying this because in hindsight I can hardly imagine anyone will ever use this. It covers the whole Latin script. Only some Asian and very exotic languages are not covered by this)

EisenbergEffect:
I think we can unify to create a single Unary AST node. Shortening the property names would be ok if the short names were meaningful still and not cryptic. I wouldn't want to shorten them and end up with something awkward. For the extension point, I'd be ok to remove that. I can't think of a lot of needs for that either. We could always add it back later if someone makes a specific request.

EisenbergEffect:
@fkleuver Looks solid :) Anything else you feel you need to do here or do you think it's all set to merge?

fkleuver:
@EisenbergEffect I must admit I'm a little nervous, but I added those unit tests for a good reason. I think they've got my back :) I'd say it's good to go.

What's the right place to add documentation for template literals and the added operators?

fkleuver:
@EisenbergEffect See latest commit. I couldn't really do one without the other .. I hope it's clear enough :)

EisenbergEffect:
@fkleuver That will do nicely :) Anything else before we merge?

fkleuver:
@EisenbergEffect Nope :) Time to press the button I think.

PR #691, opened 2018-05-27
refactor(ast/parser): remove Chain

fkleuver:
Chain isn't used anywhere and also can't be parsed from any valid expression. It appears to be a completely redundant expression.

Look only at the last commit

@bigopon


fkleuver:
@EisenbergEffect done

Issue #586, opened 2017-04-06
Unparser doesn't preserve operation precedence


huochunpeng:
**I'm submitting a bug report**

* **Library Version:**
1.2.1

**Please tell us about your environment:**
* **Operating System:**
IRRELEVANT

* **Node Version:**
IRRELEVANT

* **NPM Version:**
IRRELEVANT

* **JSPM OR Webpack AND Version**
IRRELEVANT

* **Browser:**
IRRELEVANT

* **Language:**
IRRELEVANT

**Current behavior:**
```
const parser = new Parser();
parser.parse('a&&(b||c)').toString() // 'a&&b||c' wrong logic
parser.parse('a&&(b=1)').toString()  // 'a&&b=1' syntactically wrong
parser.parse('2*(2+3)').toString() // '2*2+3' oops
```

**Expected/desired behavior:**

* **What is the expected behavior?**
Unparser should preserve operation precedence

* **What is the motivation / use case for changing the behavior?**
It's a bug to be fixed.
`unparser.spec.js` should cover this.

I tried fixing it, but failed. It looks like the unparser needs to understand the position of the expression in the tree, and also the default javascript operator precedence, in order to decide whether needs to wrap the expression in extra parenthesis.

jdanyow:
`a || (b && c)` and `a || b && c` are the same because && has higher operator precedence than ||. Same goes for the other example.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

huochunpeng:
@jdanyow sorry I was showing wrong example, try 'a&&(b||c)' will show you the bug. (I will original post later)

I am not wrong on 2nd sample 'a&&(b=1)', I think you missed it in glance, it is assignment not equal operator.

This is a valid bug.

huochunpeng:
@jdanyow One more solid error, please reopen the issue.

    parser.parse('2*(2+3)').toString() // '2*2+3' oops

jdanyow:
basically every node of type `Binary` needs to wrap it's output in parens

huochunpeng:
More than that, by looking at the operator precedence, I found `Conditional` also failed.

    parser.parse('!(true?true:true)').toString() // '(!true?true:true)' evaluated value reversed, how it kept the parens is beyond me.

There might be more.

Added extra explicit parens could solve the logic bug, but the appearance of the unparsed expression is not in minimum form.

huochunpeng:
[...]

PR #696, opened 2018-06-21
fix(parser): left-to-right associativity for nested binary expression…

fkleuver:
Somehow I missed this one. Big oops. Fixes https://github.com/aurelia/binding/issues/586#issuecomment-399044403

@EisenbergEffect

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -106,7 +106,7 @@ export class ParserImplementation {
 
     while (this.tkn & T$BinaryOp) {
       const opToken = this.tkn;
-      if ((opToken & T$Precedence) < minPrecedence) {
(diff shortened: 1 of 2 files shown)


qraynaud:
Thanks @fkleuver. You are really quick!

fkleuver:
Not quite there yet. I added some failing tests. I need to go for about a 2 hours and will fix this properly asap when im back.

fkleuver:
Agreed. Binary parsing received the biggest rework and I only tested precedence, not associativity. My sincere apologies!

qraynaud:
Thanks a lot. And really you don't need to apologize. I could have done the same. I'm just concerned about others that might be struggling right now to understand why their shiny component is broken :-). I've added parens on my side and everything is fine again.

PR #703, opened 2018-07-28
Feature: inverted (or arbitrary) checkbox

rluba:
This adds the ability to use checkboxes with inverted boolean values, i.e., checked means `false` and unchecked means `true`: 

```Html
<input type="checkbox" checked.bind="something.inverted" model.bind="false" model-unchecked.bind="true">
```

But it’s not limited to inverting the checkbox. It can also be used to toggle between arbitrary values:
```Html
<input type="checkbox" checked.bind="displayMode" model.bind="'color'" model-unchecked.bind="'gray'"> Use color
```

### DIFF
--- a/src/checked-observer.js
+++ b/src/checked-observer.js
@@ -56,24 +56,28 @@ export class CheckedObserver {
   }
 
   synchronizeElement() {
-    let value = this.value;
-    let element = this.element;
-    let elementValue = element.hasOwnProperty('model') ? element.model : element.value;
-    let isRadio = element.type === 'radio';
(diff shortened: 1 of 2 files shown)


rluba:
I haven’t updated the documentation yet. I’ll do that if this feature is accepted.

StrahilKazlachev:
What is the benefit over using a value converter?

rluba:
Good point. I was so used to the AngularJs way (`<input …  ng-true-value="false" ng-false-value="true">` that I didn’t think about using a value converter.

A value converter is probably the better solution for this problem in Aurelia. 👍

PR #704, opened 2018-08-01
fix(ArrayObservation): ensure patch applied only once

### LINKED ISSUE #699: Add protection agains double registration on Array.prototype.*
**I'm submitting a feature request**

* **Library Version:**
1.* and 2.*

* **Operating System:**
Windows 10

* **Node Version:**
9.2.1

* **NPM Version:**
6.1.0

webpack 3.12.0

* **Browser:**
all

* **Language:**
all

**Current behavior:**
Aurelia binding is [current being included in the dependency graph in two versions](https://github.com/aurelia/templating-router/issues/80). The binding library adds functionality with side effects on e.g. `Array.prototype.splice`, and when included in two versions, it gets added twice, causing all sorts of subtle bugs as change tracking events are recorded twice.

A clear error in the browser console telling that something is very wrong because side effect functionality is being double registered.

An error like this would have saved me some hours of debugging inside aurelia itself.

bigopon:
Things may still go horribly wrong if one ended up having two versions of `aurelia-binding` in the final bundle, but at least array patching won't happen twice. Also trim some unnecessary fat out

Edit: this will work as long as there is no difference between v1 and v2 `ModifyArrayObserver` / `ModifyCollectionObserver`

@EisenbergEffect Should we also replace global `karma start` with npm script ?

resolves #699

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
     "binding",
     "databinding"
   ],
+  "scripts": {
+    "test": "karma start"
+  },
   "homepage": "http://aurelia.io",
   "bugs": {
     "url": "https://github.com/aurelia/binding/issues"
--- a/src/array-observation.js
+++ b/src/array-observation.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-extend-native */
 import {ModifyCollectionObserver} from './collection-observation';
+import * as LogManager from 'aurelia-logging';
 
 const arrayProto = Array.prototype;
 const pop = arrayProto.pop;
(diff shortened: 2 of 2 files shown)


huochunpeng:
Suggest to log an error in console if it detects already patched, explain possible cause (3rd party plugin), and provide a link to suggested solution. Because duplicated aurelia-binding will inevitably introduce weird bug even without duplicated patch on Array.prototype.

bigopon:
@huochunpeng nice suggestion 👍

EisenbergEffect:
I love this idea. One change: could you switch the lookups back to using property access instead of array access? That will probably cause a perf change and since this is a hot code path, I'd rather not change that.

bigopon:
@EisenbergEffect done

EisenbergEffect:
Thanks!

rmja:
Just a reminder that this change should be merged back into v1 in a minor or patch release to work properly😁

EisenbergEffect:
working on a 1.7.2 release with this same fix

EisenbergEffect:
1.7.2 and 2.1.3 both published with this fix.

PR #714, opened 2018-09-21
docs(how-it-works): Add sentences about observed props added during binding

### LINKED ISSUE #713: One-Way adds 'undefined' property value on object with no properties
Although I am sure this is the expected behavior and my scenario might be an edge case.

* **Library Version:**
1.6.0 (aurelia-framework 1.2)

* **Operating System:**
Windows 10

* **Node Version:**
v8.9.4

* **NPM Version:**
6.3.0

webpack 3.12.0

* **Browser:**
all

* **Language:**
all

**Current behavior:**
When using `one-way` binding on an objects property in a view, that property is added to the object with a value an `undefined` when that object is created without any properties, which makes it feel like the view is updating my model (aka `two-way`).  See https://stackblitz.com/edit/aurelia-javascript-qekbtf?embed=1&file=src/app.js for an example.

The `undefined` value gets added [here](https://github.com/aurelia/binding/blob/a680c7203e850ea37b4e1147365d2c0a07e94968/src/property-observation.js#L97) 

No property is added the object if it does not exist and the value is undefined

Using the example I provided: When `canDeactivate` runs I compare the initial `foo` to the current `foo` so I can warn users of any unsaved changes before navigating away. I not only compare property values, but property count too (i am using lodash.isequal). However, when a user goes to a view model and does nothing, then tries to navigate away, they receive my warning because the properties in the view have been added to my empty object. There are many ways I can work around this, and I have done that in my app. However, in my example I am using redux so i am expecting all state mutations to come from my reducer. There might not be a solution, but I wanted to address this before I moved on since it bothered me that something outside of my reducer was changing my object and that the change originated from the view in `one-way` binding.

jmzagorski:
Aurelia Binding adds the observing property to source object when it does
not exist for a few of the binding commands. This can be confusing for
state management applications that might have a very large state and do not
define every single property on their initial state. After binding runs
in the view model, their state now has defined properties that were
not on an initial state. I think it is important to document this since
state management (like aurelia-store) are becoming popular. Plus it gives a 
little more insight into the binding logic.

I added two sentences to the end of the marked paragraph, as i thought it would be appropriate after talking about each binding commands observing logic. Fee free to change anything that might sound off or is incorrect.

closes #713


EisenbergEffect:
Looks great!

PR #720, opened 2018-10-28
WIP: Parser fixes

fkleuver:
In response to https://discourse.aurelia.io/t/horrendous-bug/1816/7

I've made some initial effort to backport the vNext parser to vCurrent. It is "working" in the sense that:
- The 13k or so tests (also backported from vNext) are all passing
- The error messages are not checked; the Reporter still needs to be replaced.

The vNext parser is more robust (with errors), thorough (with edge cases) and a little bit smaller/faster, so backporting it seems beneficial for vCurrent.

However unicode string literals and numeric exponents are dropped in vNext. Either vNext should get these again, or I just need to add that back in to the backported parser.

The vNext parser tests refused to run in the old setup, but a bunch of the vCurrent tests are failing in the "new" setup (karma+webpack) so this needs some work.
The above (and lack of const enums) makes me consider to just "do it right" and port aurelia-binding to TypeScript in its entirety. We have some older vNext branches we can copy from, so it wouldn't be nearly as much work as @bigopon had to spent on the router.

Thoughts @EisenbergEffect ?


EisenbergEffect:
I feel pretty nervous about making big changes to vCurrent at this point. Is there any way we can easily fix the bug with a smaller change?

fkleuver:
You know how I like to make things difficult :P
https://github.com/aurelia/binding/pull/721

PR #721, opened 2018-10-30
fix(parser): throw on unterminated quote instead of infinite loop

fkleuver:
@EisenbergEffect


EisenbergEffect:
I feel a lot more comfortable with this 😏

bigopon:
> I feel a lot more comfortable with this 😏

Folk: Oh look, the handle is broken.
Fred: Lets replace the house

PR #724, opened 2018-11-20
Reduce node_modules directory size

pndewit:
Apparently I am having some modules that require a different version of some aurelia modules and these reports folders are over 1MB each. There might be even more to gain in this area.

EDIT: Let me know which directories/files can also be added to this list and I will update this PR. Noticed also other Aurelia repo's can have this directory. Let me know if you need me to submit a PR there too.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 jspm_packages
 bower_components
 .idea
+build/reports


fkleuver:
That's.. interesting. Good find. We definitely don't want that stuff in the published outputs

pndewit:
Bump, is anyone looking at the impact or other directories/files that can be removed from the published output?

bigopon:
I've rarely look in to it. but the change lgtm. Pinging @EisenbergEffect

EisenbergEffect:
Frankly, I forgot this was even in there. We haven't really used this much in a long tie. We can certainly merge this. @pndewit If you want to submit similar PRs to other repos that are causing issues, we can certainly get those merged.

pndewit:
No problem, glad I could help. Creating PRs now.

StrahilKazlachev:
If we are going down that road, I *strongly* support @jods4 suggestion, from some time ago, for using the [files](https://docs.npmjs.com/files/package.json#files) field in  `package.json`.

PR #726, opened 2018-11-22
chore(build): fix shape-doc task

fkleuver:
@EisenbergEffect


EisenbergEffect:
not sure how that happen and didn't break things before...

PR #728, opened 2018-11-27
Update Collection Observer Typings

alienriver49:
Correction to the ICollectionObserverSplice interface. Adding missing methods to the InternalCollectionObserver interface and ObserverLocator class.

One thing I noticed with the InternalCollectionObserver interface was the `subscribe(callback: (changeRecords: any) => void): void;` and `unsubscribe(callback: (changeRecords: any) => void): void;` method definitions; am I crazy to think that these are not supported / depreciated?

### DIFF
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -187,8 +187,10 @@ export declare interface ICollectionObserverSplice<T = any, K = any> {
    * The type of change that has taken place. Valid options are "add", "delete", and "update".
    * 
    * *Note:* "update" is invalid for Set.
+   * 
+   * *Note:* "clear" is only valid for Map and Set.
    */
-  type: "add" | "delete" | "update";
+  type: "add" | "delete" | "update" | "clear";
 }
 
 /**
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -184,7 +184,7 @@ export declare interface ICollectionObserverSplice<T = any, K = any> {
   value: T;
                                                    
   /**
-   * The type of change that has taken place. Valid options are "add", "delete", and "update".
+   * The type of change that has taken place. Valid options are "add", "delete", "update", and  "clear".
    * 
    * *Note:* "update" is invalid for Set.
    * 
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -184,10 +184,13 @@ export declare interface ICollectionObserverSplice<T = any, K = any> {
   value: T;
                                                    
   /**
-   * The type of change that has taken place. Valid options are "add", "delete", and "update".
-   * "update" is invalid for Set.
+   * The type of change that has taken place. Valid options are "splice", "add", "delete", "update", and "clear".
+   * 
+   * *Note:* "splice" is only valid for Array; "add", "update", and "clear" are invalid for Array.
+   * 
+   * *Note:* "update" is invalid for Set.
    */
-  type: "add" | "delete" | "update";
+  type: "splice" | "add" | "delete" | "update" | "clear";
 }
 
 /**
@@ -509,6 +512,28 @@ export declare interface InternalCollectionObserver {
    * @param callable A callable object.
    */
   unsubscribe(context: any, callable: Callable): void;
+  /**
+   * Adds a change record to the collection observer.
+   * @param changeRecord 
+   */
+  addChangeRecord(changeRecord: ICollectionObserverSplice): void;
+  /**
+   * This will flush the change records of this observer and call any subscribers if applicable.
+   */
+  flushChangeRecords(): void;
+  /**
+   * Reset the observer to the passed collection and call any subscribers with changes between the current collection and the reset collection.
+   * @param oldCollection 
+   */
+  reset(oldCollection): void;
+  /**
+   * Get a length observer for this collection.
+   */
+  getLengthObserver(): any;
+  /**
+   * This will call subscribers notifying of changed records.
+   */
+  call(): void;
 }
 
 /**
@@ -821,6 +846,10 @@ export declare class ObserverLocator {
    * Gets an observer for map mutation.
    */
   getMapObserver(map: Map<any, any>): InternalCollectionObserver;
+  /**
+   * Gets an observer for set mutation.
+   */
+  getSetObserver(set: Set<any>): InternalCollectionObserver;
 }
 
 /**
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -145,7 +145,7 @@ export declare interface CollectionObserver {
 /**
  * The change record of a collection mutation. 
  */
-export declare interface ICollectionObserverSplice<T = any> {
+export declare interface ICollectionObserverSplice<T = any, K = any> {
   /* ArrayObserverSplice */
   /**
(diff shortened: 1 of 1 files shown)


src/aurelia-binding.d.ts
```diff
@@ -184,10 +184,13 @@ export declare interface ICollectionObserverSplice<T = any> {
...
-   * The type of change that has taken place. Valid options are "add", "delete", and "update".
-   * "update" is invalid for Set.
+   * The type of change that has taken place. Valid options are "splice", "add", "delete", "update", and "clear".
+   * 
+   * *Note:* "splice" is only valid for Array; "add", "update", and "clear" are invalid for Array.
+   * 
+   * *Note:* "update" is invalid for Set.
    */
-  type: "add" | "delete" | "update";
+  type: "splice" | "add" | "delete" | "update" | "clear";
```
fkleuver: What's not so obvious from the code is that "splice" is actually normalized into "add"s and "delete"s, and "clear" is normalized into "delete"s. The original documentation is correct here for the end-users perspective. You will never receive "splice" or "clear" operations from this code.

src/aurelia-binding.d.ts
```diff
@@ -509,6 +512,28 @@ export declare interface InternalCollectionObserver {
    * @param callable A callable object.
    */
   unsubscribe(context: any, callable: Callable): void;
+  /**
+   * Adds a change record to the collection observer.
+   * @param changeRecord 
+   */
+  addChangeRecord(changeRecord: ICollectionObserverSplice): void;
```
fkleuver: This isn't necessarily meant to be public API. It requires rather intricate knowledge of the internal data structures and algorithms to be able to use this without messing things up.

src/aurelia-binding.d.ts
```diff
@@ -509,6 +512,28 @@ export declare interface InternalCollectionObserver {
...
+  /**
+   * This will flush the change records of this observer and call any subscribers if applicable.
+   */
+  flushChangeRecords(): void;
```
fkleuver: I guess this can be exposed since it serves a similar purpose to `flushMicroTaskQueue`, and I can see potential legitimate use cases for this.

src/aurelia-binding.d.ts
```diff
@@ -509,6 +512,28 @@ export declare interface InternalCollectionObserver {
...
+  /**
+   * Reset the observer to the passed collection and call any subscribers with changes between the current collection and the reset collection.
+   * @param oldCollection 
+   */
+  reset(oldCollection): void;
+  /**
+   * Get a length observer for this collection.
+   */
+  getLengthObserver(): any;
```
fkleuver: This seems OK. It can make sense for a user to want the length observer, albeit very rarely.

src/aurelia-binding.d.ts
```diff
@@ -509,6 +512,28 @@ export declare interface InternalCollectionObserver {
...
+  /**
+   * This will call subscribers notifying of changed records.
+   */
+  call(): void;
```
fkleuver: I don't think this should be exposed as it's a special interface for the task queue to be called. Especially when there is already the other mechanism to flush. Would just confuse users, I think.

src/aurelia-binding.d.ts
```diff
@@ -821,6 +846,10 @@ export declare class ObserverLocator {
    * Gets an observer for map mutation.
    */
   getMapObserver(map: Map<any, any>): InternalCollectionObserver;
+  /**
+   * Gets an observer for set mutation.
+   */
+  getSetObserver(set: Set<any>): InternalCollectionObserver;
```
fkleuver: Nice find, this definitely belongs here

src/aurelia-binding.d.ts
```diff
@@ -145,7 +145,7 @@ export declare interface CollectionObserver {
 /**
  * The change record of a collection mutation. 
  */
-export declare interface ICollectionObserverSplice<T = any> {
+export declare interface ICollectionObserverSplice<T = any, K = any> {
```
fkleuver: Good catch

fkleuver:
> One thing I noticed with the InternalCollectionObserver interface was the subscribe(callback: (changeRecords: any) => void): void; and unsubscribe(callback: (changeRecords: any) => void): void; method definitions; am I crazy to think that these are not supported / depreciated?

Definitely not crazy, this is a bit confusing. I'm guessing you're looking at `subscribe(context, callable)`. That contract is misleading. What you should read it like is this overload:

```ts
subscribe(callback);
subscribe(context, callback);
subscribe(contextOrCallback, callback?)
```

The change handler would match the first signature, the callback. So it's correct.

src/aurelia-binding.d.ts
alienriver49: Interesting! That is good to know, I will revert this change back to the original.

src/aurelia-binding.d.ts
alienriver49: I completely understand, I can remove this addition.

src/aurelia-binding.d.ts
alienriver49: Awesome!

src/aurelia-binding.d.ts
alienriver49: Okay cool! Are you good with reset being exposed?

src/aurelia-binding.d.ts
alienriver49: I agree, I will revert the addition of this.

alienriver49:
Thanks for the clarity on the `subscribe` and `unsubscribe`, makes perfect sense. I've reverted the exposing of the functionality which shouldn't be public, let me know if there is anything else!

src/aurelia-binding.d.ts
alienriver49: I actually just checked this for maps and sets and am receiving a type of clear when the clear method is called. Let me see if I can get a Gist going with this.

alienriver49: Here's my Gist showing this: https://gist.run/?id=43a9adb8ba9004936e710bfef7269af4

fkleuver: Huh. I stand corrected! And splice?

fkleuver:
@bigopon I'm inclined to say this is good to go but since this is vCurrent I'd like to be a bit extra careful with public API. Can you do a quick sanity check here as well?

src/aurelia-binding.d.ts
bigopon: No splice is correct I think https://gist.run/?id=afc8391c38d90dd5a398ed40c1698790

bigopon: In the gist, it shows that 'clear' is a valid type for Map | Set

bigopon:
I think it's safe to add `"clear"`, with a note that `"clear"` is only valid for Map / Set, unless we want to use more generic conditional type, which probably not a good idea, since there could be app running in prod with typescript older version.

src/aurelia-binding.d.ts
alienriver49: Awesome! Works for me, I'll re-add clear and make a note. Thanks guys!

Alexander-Taran:
and the winner is?

fkleuver:
@EisenbergEffect Are you cool with the additions to the public api?

EisenbergEffect:
Thanks everyone for the attention to detail here!

Issue #134, opened 2015-07-09
empty @bindable brings error in babel runtime


prefinnx:
Hi,  I am playing with aurelia navigation skeleton, and try to load ES6(7) modules directly from the 'src'  directory, then got  and error and page is not loaded. I've made some effort trying to solve but failed. 

So, is here a clever trick that I can use to get it working? I really enjoy  working with ES6 code directly with SystemJS  :D

 I dig into it and get this line in `nav-bar.js`
```javascript
@bindable router = null;
```
I also test to not use the nav-bar in `app.html` then Aurelia is loading right: 

```html
  <require from='./nav-bar'></require>
  <nav-bar router.bind="router"></nav-bar>
```
to
```html
  <nav class="navbar" role="navigation">
        <span>${router.title}</span>
  </nav>
```
##env
- npm:babel-runtime 5.1.13
- github:aurelia/framework 0.11.0
- github:aurelia/binding 0.6.1
- FF 39.0
- Darwin Kernel Version 14.1.0

##error msg
```
GET http://localhost:9001/src/app.html 200 OK 12ms	HTMLImp....min.js (line 11)
DEBUG [templating] importing resources for app.html ["nav-bar"] index.js (line 41)
GET http://localhost:9001/src/nav-bar.js 200 OK 2ms	es6-mod...ader.js (line 7)
GET http://localhost:9001/jspm_packages/npm/babel-run.../helpers/define-decorated-property-descriptor.js 200 OK 10ms	es6-mod...ader.js (line 7)
GET http://localhost:9001/jspm_packages/npm/babel-runtime@5.1.13/helpers/create-decorated-class.js 200 OK 10ms	es6-mod...ader.js (line 7)
GET http://localhost:9001/src/nav-bar.html 200 OK 2ms	HTMLImp....min.js (line 11)
DEBUG [templating] importing resources for nav-bar.html [] index.js (line 41)
Unhandled promise rejection
Error: Error instantiating NavBar.
 ------------------------------------------------ inner error:
exports.default@http://localhost:9001/jspm_packages/npm/babel-runtime@5.1.13/helpers/define-decorated-property-descriptor.js:9:22
NavBar@http://localhost:9001/src/nav-bar.js!eval:23:11
construct@http://localhost:9001/jspm_packages/npm/core-js@0.9.7/modules/es6.reflect.js:47:18
invoke@http://localhost:9001/jspm_packages/github/aurelia/dependency-injection@0.7.1/metadata.js:183:18
[...]

PR #729, opened 2019-01-11
feat(connect-queue) optionally change the connect queue threshold

dannyBies:
As discussed in [#134](https://github.com/aurelia/templating-binding/issues/134) I implemented the suggested feature and added a test to verify this behaviour.
After experimenting a bit more I have been able to get performance gains of around 60% on the initial loading of my screens by disabling the connect-queue.

Note:
I have added enableConnectQueue to make sure that the connectQueue can be enabled again after calling disableConnectQueue(). Technically this can be achieved by setConnectQueueThreshold(100) but I don't think developers should have to know that the default is 100. I can remove disableConnectQueue if needed.


fkleuver:
Excellent, thanks for taking the initiative and also adding a test :)

Just one request. Would you mind also creating a test that verifies `enableConnectQueue` and `disableConnectQueue`? I know it may seem a bit redundant/obvious, but I would just like to make sure that 1) it's covered and 2) tests protect against regressions in the future should someone touch this code again for one reason or another. It's one of the things that should help vCurrent to keep working and stay robust even in the future.

Other than that, looks solid!

dannyBies:
@fkleuver I've added the testcases. Unfortunately this has resulted in flaky tests. With every test run 2 or 3 tests randomly fail with the message Expected 'bar' to be 'baz'. I'm not familiar enough with the codebase to tell if this has to do with the implementation or the tests. 

I've tried to flush the TaskQueue and call flush in connect-queue.js and tried to set immediate to 0, nothing seems to help. Do you have any ideas what the problem could be?

fkleuver:
There are probably some race conditions going on with other tests because rAF has a bigger delay than any other kind of promise or timeout. You could try forcing the test to both start running after the previous rAF resolved and only finish after the next rAF resolved to ensure there are no queue leftovers from either earlier or later tests.

You could try something like this:

```ts
it('works?', done => {
  PLATFORM.requestAnimationFrame(() => {
    // test code

    PLATFORM.requestAnimationFrame(() => {
      done();
    });
  });
});
```

If that's still flaky, at least that gives you the appropriate timings to be trying stuff like manually flushing and whatnot. `setTimeout` is definitely not going to accomplish a whole lot :)

dannyBies:
That seemed to have worked, thanks for your help!
Are there any other tests you think I should add or is it okay like this?

fkleuver:
Nice, thank *you* for fixing this!
I don't think additional tests are needed. The assertions are very precise so this looks solid both for verifying the behavior and for preventing regressions.
LGTM @EisenbergEffect

EisenbergEffect:
Nice. Tests and all!

dannyBies:
Thanks for merging this @EisenbergEffect!  When do you think this will be released?

EisenbergEffect:
I've got a slew of new releases for the various libraries that are on my plate to get out. I think I'll probably start working on getting those out tomorrow. Not sure if `binding` will be in tomorrow's batch or not. It will depend on how the process goes. It should be this week though.

Side note: Our vCurrent release process isn't as fast as we'd like but we've done massive work for vNext to make this better. We're actually able to do nightly builds for vNext now and with a click of a button I can turn any of those into an official release. So, speed of release is something we've worked on, just much easier to do with our new monorepo than with our vCurrent multi-repo approach. So, thanks for your patience.

PR #731, opened 2019-01-17
fix(PropertyObservation): use setAttribute in property accessor if target is HTMLElement

### LINKED ISSUE #730: propertyAccessor not triggering DOM MutationObserver for custom attributes
I have a view with a select element:
```html
<template>
  <select ref="selectElement" value.bind="selectedValue">
    <option repeat.for="opt of options" model.bind="opt.key">${opt.name}</option>
  </select>
</template>
```
I overlay this select element with a bootstrap selectpicker. This picker needs to be refreshed whenever the underlying select element changes (i.e. options updated etc). For this purpose, I use a DOM MutationObserver on the select element like:
```ts
this.optionsChangeObserver = DOM.createMutationObserver(mutations => {
  this.taskQueue.queueMicroTask(() => {
	$(this.selectElement).selectpicker('refresh');
  });
});

this.optionsChangeObserver.observe(this.selectElement, {
  childList: true,
  subtree: true,
  attributes: true,
  attributeFilter: ['value', 'model']
});
```
The issue is, that any update to `opt.key` will be applied to the target `option` DOM element by the aurelia-binding propertyAccessor:
```js
var propertyAccessor = exports.propertyAccessor = {
  getValue: function getValue(obj, propertyName) {
    return obj[propertyName];
  },
  setValue: function setValue(value, obj, propertyName) {
    obj[propertyName] = value;
  }
};
```
This triggers an attribute change event for the DOM MutationObserver as long as `propertyName` is not a custom attribute (see https://stackoverflow.com/a/21044802/4005051). For the custom `model` attribute, no attribute change is triggered for the MutationObserver and hence, the selectpicker is not refreshed in this example. The whole thing works as soon as I use the standard `value` attribute vor the `option` elements but then I have the known issues with non-string values - the reason for Aurelia's support for the `model` attribute.

One potential fix for this is to check if the target object is a HTMLElement in the propertyAccessor and to use `setAttribute` in this case:
```js
var propertyAccessor = exports.propertyAccessor = {
  getValue: function getValue(obj, propertyName) {
    return obj[propertyName];
  },
  setValue: function setValue(value, obj, propertyName) {
    if (obj instanceof HTMLElement) {
      obj.setAttribute(propertyName, value);  
    } else {
      obj[propertyName] = value;  
    }
  }
};
```
With this change, the example works as expected with the `model` attribute.

Mobe91:
fixes #730

### DIFF
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -5,7 +5,13 @@ const logger = LogManager.getLogger('property-observation');
 
 export const propertyAccessor = {
   getValue: (obj, propertyName) => obj[propertyName],
-  setValue: (value, obj, propertyName) => { obj[propertyName] = value; }
+  setValue: (value, obj, propertyName) => {
+    if (obj instanceof HTMLElement) {
+      obj.setAttribute(propertyName, value);
+    } else {
+      obj[propertyName] = value;
+    }
+  }
 };
 
 export class PrimitiveObserver {


bigopon:
@Mobe91 Thanks for the suggestion, but it's working as expected. By default, Aurelia works with **properties** instead of attributes, which mean `model.bind=...` means `.model = ...`

What you can do is to create an additional binding to reflect select value to the DOM attribute, using `attr` behavior:

```html
<select value.to-view="selectedValue & attr">
```

I have created an example sandbox here https://codesandbox.io/s/n3r28375l . Can you have a check if this help with the issues you are having. Btw, if you are to observe `model` attribute on `<option />` elements, I don't think you can observe it from `<select/>` element.

Mobe91:
I guess you saved my day, awesome! 😃 
I did not know about the `attr` binding behavior but now I see that it is even mentioned on Stackoverflow: https://stackoverflow.com/a/38574417/4005051

Closing this PR

PR #735, opened 2019-01-24
fix(event-manager) prevent double subscribing/disposing

Alexander-Taran:
I've stumbled upon a possibility to call dispose() more than one time on an Subscription returned by EventManager
that does not unsubscribe twice, but does decrease internal counter
It is also possible to subscribe to one event more than one time. Last callback wins.

This PR fixes counter fluctuations and provides errors in console for developers.
I wanted not to introduce breaking changes, but to increase awareness.
So we all can still shoot ourselves in the foot, but we will be warned now.


fkleuver:
Overall it looks good I think. Nice little bit of extra robustness. Would be nice though if the PR can at least be coverage neutral. Should go up not down in vcurrent :)
@bigopon can you take a look as well?

src/event-manager.js
```diff
@@ -104,9 +106,10 @@ class DelegateHandlerEntry {
   }
 
   decrement() {
-    this.count--;
-
     if (this.count === 0) {
+      emLogger.warn('The same EventListener was disposed multiple times.');
+    }
+    if (--this.count === 0) {
```
bigopon: This and the block following it don't seem to guard against over disposal anymore, is this intentional?

src/event-manager.js
```diff
@@ -173,8 +178,13 @@ class DefaultEventStrategy {
       delegatedHandlers = this.delegatedHandlers;
       handlerEntry = delegatedHandlers[targetEvent] || (delegatedHandlers[targetEvent] = new DelegateHandlerEntry(targetEvent));
       let delegatedCallbacks = target.delegatedCallbacks || (target.delegatedCallbacks = {});
+      if (delegatedCallbacks[targetEvent]) {
+        emLogger.error('already have a callback for event', {event: targetEvent, callback: callback});
```
bigopon: Maybe add some note here that logging is only for debugging purpose, log content can be changed at any time so there won't be any accidental breakage

src/event-manager.js
Alexander-Taran: I think I missed the else part (-:

EisenbergEffect:
@fkleuver Do we need to port this over to vNext or has that diverged too much?

fkleuver:
Good point actually. Which made me realize there's a whole other battery of tests I should be adding for these kinds of scenarios.

Alexander-Taran:
I definitely would like to have safeguards around disposing the same subscription twice
I would even go as far as delete from inside .dispose() if possible
this should throw

src/event-manager.js
```diff
@@ -173,8 +177,13 @@ class DefaultEventStrategy {
       delegatedHandlers = this.delegatedHandlers;
       handlerEntry = delegatedHandlers[targetEvent] || (delegatedHandlers[targetEvent] = new DelegateHandlerEntry(targetEvent));
       let delegatedCallbacks = target.delegatedCallbacks || (target.delegatedCallbacks = {});
+      if (delegatedCallbacks[targetEvent]) {
+        emLogger.warn('Overriding previous callback for event listener', {event: targetEvent, callback: callback, previousCallback: delegatedCallbacks[targetEvent]});
+      }
 
-      handlerEntry.increment();
+      if (!delegatedCallbacks[targetEvent]) {
```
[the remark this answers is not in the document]
Alexander-Taran: logically they are different pieces of logic
But I see your point.
We could save one comparison.

test/event-manager.spec.js
```diff
@@ -208,6 +208,24 @@ describe('EventManager', () => {
...
+      expect(delegateBubblingCountAfterDoubleSubscription).toEqual(delegateBubblingCount,"allows double subscription for bubbling");
+
+      let delegateCaptureCount = em.defaultEventStrategy.delegatedHandlers['delegate'].count;
+     
+      handlers.push(
+        em.addEventListener(one, 'delegate', oneCapture, delegationStrategy.capturing, true),
+        em.addEventListener(three, 'delegate', threeCapture, delegationStrategy.capturing, true),
+      )
+      let delegateCaptureCountAfterDoubleSubscription = em.defaultEventStrategy.capturedHandlers['delegate'].count;
+      expect(delegateCaptureCountAfterDoubleSubscription).toEqual(delegateCaptureCount,"allows double subscription for capture");
```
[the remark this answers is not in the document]
Alexander-Taran: That would be above my paygrade (-:

test/event-manager.spec.js
```diff
@@ -192,6 +193,8 @@ describe('EventManager', () => {
       const oneCapture = jasmine.createSpy('one-capture');
       const threeCapture = jasmine.createSpy('three-capture');
 
+      spyOn(LogManager.Logger.prototype, 'warn')
```
fkleuver: You should do this in a `beforeEach` along with storing the original `warn` method in a local variable and restoring it in `afterEach`, otherwise the spy stays for the remainder of the tests.

Alexander-Taran: as per 
https://github.com/jasmine/jasmine/issues/1281#issuecomment-286831790
it will be cleaned automatically
and there is no way to manually restore AFAIK

fkleuver: Well, up to you. My suggestion would be this:

```ts
      let originalWarn;
      beforeEach(() => {
        originalWarn = LogManager.Logger.prototype.warn;
        spyOn(LogManager.Logger.prototype, 'warn')
      });
      afterEach(() => {
        LogManager.Logger.prototype.warn = originalWarn;
      });
```

Alexander-Taran: done

fkleuver:
Yeah, my WIP PR is high prio for me atm so it's better if I don't try to remember all the little things.

Alexander-Taran:
It is worth noting as well that currently we do not have a way to have more than one callback subscribed to an event per element.
Should we consider it? @EisenbergEffect  @fkleuver

fkleuver:
@Alexander-Taran What would be the API / templating syntax?

EisenbergEffect:
I'll follow up by creating an issue to track this in vNext.

Alexander-Taran:
@EisenbergEffect I added it to Mother issue.

EisenbergEffect:
Thanks @Alexander-Taran !

Alexander-Taran:
@fkleuver templating - don't know.. it does not seem to be relevant there. 
but through EventManager API you could subscribe to element/event pair more than one time. 
Nothing prevents nor it should.
But as it is now - only the last callback will be called.
Either we should declare it as not supported and prevent or figure out a performant way to allow.

PR #740, opened 2019-02-27
fix(build): don't build declaration

bigopon:
fix build issue where build process for system and native-modules would produce typing file and override hand-authored typing file

@fkleuver @EisenbergEffect


EisenbergEffect:
Does this need to go against the develop branch?

bigopon:
@EisenbergEffect Yes, fixed

PR #741, opened 2019-03-02
fix(name-binding): add extra props to update one time mode

bigopon:
Thanks to `@edwrd01` on Discourse at https://discourse.aurelia.io/t/issue-using-ref-in-a-repeat-for-with-a-value-converter/2328/5 , we uncover an issue related the combination of `ref` binding, `repeat` and `value-converter/binding-behavior`. The issue is due to `repeat` custom attribute unable to update ref binding, in case of a mutation in source collection, piped through value converter / binding behavior. i.e:

```html
  <div repeat.for="item of items | vc & bb">
```

As it uses this https://github.com/aurelia/templating-resources/blob/0896552d553af0d69399b9b227317a26bdc14005/src/repeat-utilities.js#L109-L115 under the hood to update the bindings that do not employ observation, `NameBinder` is one of those binding types. There are two ways to fix it:

1. add method `updateOneTimeBindings` to reassign target
2. add binding mode `oneTime`, and add method `call` to reassign target

I went with (2) since it seems to align better with other bindings, also the location of this class (binding module).

cc @EisenbergEffect @fkleuver

### DIFF
--- a/src/name-expression.js
+++ b/src/name-expression.js
@@ -74,6 +74,9 @@ class NameBinder {
   }
 
   call() {
+    if (!this.isBound) {
+      return;
+    }
     this.sourceExpression.assign(this.source, this.target, this.lookupFunctions);
   }
 
--- a/src/name-expression.js
+++ b/src/name-expression.js
@@ -1,3 +1,5 @@
+import { bindingMode } from './binding-mode';
+
 function getAU(element) {
   let au = element.au;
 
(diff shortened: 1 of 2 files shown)


EisenbergEffect:
This works for me. Let's make sure @fkleuver has a look and also considers this for vNext so we don't get a regression there after this is addressed in vCurrent.

fkleuver:
This makes good sense to me. This would translate to implementing `handleChange` in `ref.ts`.

Just a friendly reminder, you could use this repo for e2e / integration tests: https://github.com/aurelia/e2e-tests
If you want to integration test this with real repeater / aurelia app, just add some tests there and make another push here, then CircleCI will run those tests against the binding version of this PR (and the `npm@latest` of all the other repos)

fkleuver:
@bigopon do we have a matching issue for this in vNext so I/we don't forget?

bigopon:
Oh, we don't. Will add

PR #743, opened 2019-03-15
fix(connectable-binding): remove the silent limit of 100 observers per expression

### LINKED ISSUE #742: Infinite loop with 100+ observers

* **Library Version:**
2.2.0

* **Operating System:**
All1|10]

* **Node Version:**
11.9.0

* **NPM Version:**
6.9.0

none

* **Browser:**
all

* **Language:**
all

**Current behavior:**
Aurelia ends in infinite loop when adding more than 100 observers for one expression.
This happens because the while connecting an expression an observer is created and an observer property name is searched. The current logic in front of creates an array with slot names _observer0-99. Later a while loop increases a counter and checks if there is a slot name unused yet. At the end it results in slotname "undefined" all the time and the infinite loop reached.

https://github.com/aurelia/binding/blob/a680c7203e850ea37b4e1147365d2c0a07e94968/src/connectable-binding.js#L22

Aurelia allows adding more than 100 observers or throws an exception which explains why it is not allowed. In my eyes it should be up to the user to decide for that.

To be honest I've no clue why you created the logic like that but you may had some thoughts on that. I could create a PR to generate the property name on the fly and allow more than 100 observers but I'm not sure if that would be good.

Also I have to say that it raises a bug in our system preventing users to open details page with big entity.

fkleuver:
This is directly ported from [vNext](https://github.com/aurelia/aurelia/blob/master/packages/runtime/src/binding/connectable.ts#L15)

fixes #742

### DIFF
--- a/src/connectable-binding.js
+++ b/src/connectable-binding.js
@@ -1,12 +1,20 @@
 import {sourceContext} from './call-context';
 
+
 const slotNames = [];
 const versionSlotNames = [];
-
-for (let i = 0; i < 100; i++) {
-  slotNames.push(`_observer${i}`);
-  versionSlotNames.push(`_observerVersion${i}`);
+let lastSlot = -1;
+function ensureEnoughSlotNames(currentSlot) {
+  if (currentSlot === lastSlot) {
+    lastSlot += 5;
+    const ii = slotNames.length = versionSlotNames.length = lastSlot + 1;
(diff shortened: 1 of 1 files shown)


fragsalat:
👍

fkleuver:
Re-targeting this to the develop branch

PR #744, opened 2019-03-18
fix(build): run build-dts after all others

bigopon:
In my previous attempt at #740, `declaration` in `tsconfig` was changed to false to make build not override type definition file in dist. But proper fix is to run `build-dts` task after all other tasks, so it would always make sure we have right type definitions in distribution

@EisenbergEffect @fkleuver

### DIFF
--- a/build/tasks/build.js
+++ b/build/tasks/build.js
@@ -107,8 +107,8 @@ gulp.task('build', function(callback) {
     'clean',
     'build-index',
     compileToModules
-      .map(function(moduleType) { return 'build-babel-' + moduleType })
-      .concat(['build-dts']),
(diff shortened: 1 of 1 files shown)


fkleuver:
Thanks for fixing this @bigopon

fkleuver:
@bigopon Just 2 things for next time:
- If fixing build stuff, please use the `build:` / `ci:` or `chore:` prefix as they don't end up in the change log. This is not really a fix for binding library itself
- Please target the develop branch :) I duplicated your PR [here](https://github.com/aurelia/binding/commit/28e2d77620c4972ffcfab395c137dfa87e71984e)

PR #757, opened 2019-07-06
fix(typing): Make parentOverrideContext optional

### LINKED ISSUE #756: Infinite loop in aurelia-bindingd.ts OverrideContext

* **Library Version:**
"version": "2.3.0"

**Current behavior:**
```ts
export declare interface OverrideContext {
  parentOverrideContext: OverrideContext;
  bindingContext: any;
}
```

parentOverrideContext type makes an infinite loop.

parentOverrideContext should be optional, while debugging an actual binding the value is null.

`parentOverrideContext?: OverrideContext;`

Expecting properly generated typings

This typing files makes it impossible to properly mock a binding. I now am forced to locally overwrite the typings.

**Work around**:
Add a local aurelia-binding.d.ts with the following content:
```ts
export * from 'aurelia-binding';

declare module 'aurelia-binding' {
    export declare interface OverrideContext {
        parentOverrideContext?: OverrideContext;
        bindingContext: any;
    }
}
```

michaelw85:
Make property optional to prevent infinite loop in typing.

Closes #756

### DIFF
--- a/src/aurelia-binding.d.ts
+++ b/src/aurelia-binding.d.ts
@@ -6,7 +6,7 @@ import { Container } from 'aurelia-dependency-injection';
  * object take precedence over members of the bindingContext object.
  */
 export declare interface OverrideContext {
-  parentOverrideContext: OverrideContext;
+  parentOverrideContext?: OverrideContext;
   bindingContext: any;
 }
 


michaelw85:
Build is broken, seems unrelated to my change.
❌ `INFO [Bundle] Error: ENOENT: no such file or directory, open 'node_modules/@babel/polyfill/browser.js'`

PR #758, opened 2019-07-16
fix(event-manager): shadow dom event bubbling

### LINKED ISSUE #755: Delegate event bubbling does not match native behavior in shadow dom

* **Library Version:**
Latest version, current master (2.3.2)

* **Operating System:**
Windows 10

* **Node Version:**
v10.15.1

* **NPM Version:**
6.9.0

* **Webpack**
4.27.0

* **Browser:**
Chrome Version 75.0.3770.100 (Official Build) (64-bit)

* **Language:**
all

**Current behavior:**
When using a click delegate on a custom element with shadow dom the callback is not fired.
```ts
<custom-comp click.delegate="expectToTrigger()"></...>
```

According to [this documentation](https://developers.google.com/web/fundamentals/web-components/shadowdom#events) and by testing in chrome click events should be able to cross the shadow dom boundary. 

Click delegate should fire on custom element regardless of using shadow dom or not.

Behavior should be similar/equal to native behavior since I'm under the impression this is the vision of Aurelia.
From a technical perspective I am now forced to use the less performant trigger or custom events.

**Possible root cause:**
[event-manager:70](https://github.com/aurelia/binding/blame/master/src/event-manager.js)
```ts
function handleDelegatedEvent(event) {
  event.propagationStopped = false;
  let target = findOriginalEventTarget(event);

  while (target && !event.propagationStopped) {
    if (target.delegatedCallbacks) {
      let callback = target.delegatedCallbacks[event.type];
      if (callback) {
        if (event.stopPropagation !== stopPropagation) {
          event.standardStopPropagation = event.stopPropagation;
          event.stopPropagation = stopPropagation;
        }
        if ('handleEvent' in callback) {
          callback.handleEvent(event);
        } else {
          callback(event);
        }
      }
    }

    // 🔽🔽🔽 this will receive null when the shadow root is reached and will break the loop
    target = target.parentNode;
  }
}
```

michaelw85:
Delegate events should be able to pass shadow dom boundry similar to native
events.

Closes #755

### DIFF
--- a/dist/es2015/aurelia-binding.js
+++ b/dist/es2015/aurelia-binding.js
@@ -3085,7 +3085,7 @@ let DelegateHandlerEntry = class DelegateHandlerEntry {
     this.eventManager = eventManager;
   }
 
-  handleDelegatedEvent(event) {
+  handleEvent(event) {
     event.propagationStopped = false;
     let target = findOriginalEventTarget(event);
 
@@ -3106,7 +3106,7 @@ let DelegateHandlerEntry = class DelegateHandlerEntry {
       }
 
       const parent = target.parentNode;
-      const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent && parent instanceof ShadowRoot;
+      const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent instanceof ShadowRoot;
 
       target = shouldEscapeShadowRoot ? parent.host : parent;
     }
@@ -3116,7 +3116,7 @@ let DelegateHandlerEntry = class DelegateHandlerEntry {
     this.count++;
 
     if (this.count === 1) {
-      DOM.addEventListener(this.eventName, this.handleDelegatedEvent, false);
+      DOM.addEventListener(this.eventName, this, false);
     }
   }
 
@@ -3124,7 +3124,7 @@ let DelegateHandlerEntry = class DelegateHandlerEntry {
     if (this.count === 0) {
       emLogger.warn('The same EventListener was disposed multiple times.');
     } else if (--this.count === 0) {
-      DOM.removeEventListener(this.eventName, this.handleDelegatedEvent, false);
+      DOM.removeEventListener(this.eventName, this, false);
     }
   }
 };
--- a/dist/es2015/aurelia-binding.js
+++ b/dist/es2015/aurelia-binding.js
@@ -1,4 +1,4 @@
-var _dec, _dec2, _class, _dec3, _class2, _dec4, _class3, _dec5, _class6, _dec6, _class8, _dec7, _class9, _dec8, _class10, _dec9, _class11, _class12, _temp, _dec10, _class13, _class14, _temp2;
+var _dec, _dec2, _class, _dec3, _class2, _dec4, _class3, _dec5, _class5, _dec6, _class7, _dec7, _class8, _dec8, _class9, _dec9, _class10, _class11, _temp, _dec10, _class12, _class13, _temp2;
 
 import * as LogManager from 'aurelia-logging';
 import { PLATFORM, DOM } from 'aurelia-pal';
@@ -3080,36 +3080,36 @@ let CapturedHandlerEntry = class CapturedHandlerEntry {
 };
 let DelegateHandlerEntry = class DelegateHandlerEntry {
   constructor(eventName, eventManager) {
-    this.handleDelegatedEvent = event => {
-      event.propagationStopped = false;
-      let target = findOriginalEventTarget(event);
-
-      while (target && !event.propagationStopped) {
-        if (target.delegatedCallbacks) {
-          let callback = target.delegatedCallbacks[event.type];
-          if (callback) {
-            if (event.stopPropagation !== stopPropagation) {
-              event.standardStopPropagation = event.stopPropagation;
-              event.stopPropagation = stopPropagation;
-            }
-            if ('handleEvent' in callback) {
-              callback.handleEvent(event);
-            } else {
-              callback(event);
-            }
-          }
-        }
-
-        const parent = target.parentNode;
-        const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent && parent instanceof ShadowRoot;
-
-        target = shouldEscapeShadowRoot ? parent.host : parent;
-      }
-    };
-
     this.eventName = eventName;
     this.count = 0;
     this.eventManager = eventManager;
+  }
+
+  handleDelegatedEvent(event) {
+    event.propagationStopped = false;
+    let target = findOriginalEventTarget(event);
+
+    while (target && !event.propagationStopped) {
+      if (target.delegatedCallbacks) {
+        let callback = target.delegatedCallbacks[event.type];
+        if (callback) {
+          if (event.stopPropagation !== stopPropagation) {
+            event.standardStopPropagation = event.stopPropagation;
+            event.stopPropagation = stopPropagation;
+          }
+          if ('handleEvent' in callback) {
+            callback.handleEvent(event);
+          } else {
+            callback(event);
+          }
+        }
+      }
+
+      const parent = target.parentNode;
+      const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent && parent instanceof ShadowRoot;
+
+      target = shouldEscapeShadowRoot ? parent.host : parent;
+    }
   }
 
   increment() {
@@ -3396,7 +3396,7 @@ export let DirtyChecker = class DirtyChecker {
   }
 };
 
-export let DirtyCheckProperty = (_dec5 = subscriberCollection(), _dec5(_class6 = class DirtyCheckProperty {
+export let DirtyCheckProperty = (_dec5 = subscriberCollection(), _dec5(_class5 = class DirtyCheckProperty {
   constructor(dirtyChecker, obj, propertyName) {
     this.dirtyChecker = dirtyChecker;
     this.obj = obj;
@@ -3437,7 +3437,7 @@ export let DirtyCheckProperty = (_dec5 = subscriberCollection(), _dec5(_class6 =
       this.dirtyChecker.removeProperty(this);
     }
   }
-}) || _class6);
+}) || _class5);
 
 const logger = LogManager.getLogger('property-observation');
 
@@ -3471,7 +3471,7 @@ export let PrimitiveObserver = class PrimitiveObserver {
   unsubscribe() {}
 };
 
-export let SetterObserver = (_dec6 = subscriberCollection(), _dec6(_class8 = class SetterObserver {
+export let SetterObserver = (_dec6 = subscriberCollection(), _dec6(_class7 = class SetterObserver {
   constructor(taskQueue, obj, propertyName) {
     this.taskQueue = taskQueue;
     this.obj = obj;
@@ -3541,7 +3541,7 @@ export let SetterObserver = (_dec6 = subscriberCollection(), _dec6(_class8 = cla
       logger.warn(`Cannot observe property '${this.propertyName}' of object`, this.obj);
     }
   }
-}) || _class8);
+}) || _class7);
 
 export let XLinkAttributeObserver = class XLinkAttributeObserver {
   constructor(element, propertyName, attributeName) {
@@ -3672,7 +3672,7 @@ export let StyleObserver = class StyleObserver {
   }
 };
 
-export let ValueAttributeObserver = (_dec7 = subscriberCollection(), _dec7(_class9 = class ValueAttributeObserver {
+export let ValueAttributeObserver = (_dec7 = subscriberCollection(), _dec7(_class8 = class ValueAttributeObserver {
   constructor(element, propertyName, handler) {
     this.element = element;
     this.propertyName = propertyName;
@@ -3721,12 +3721,12 @@ export let ValueAttributeObserver = (_dec7 = subscriberCollection(), _dec7(_clas
       this.handler.dispose();
     }
   }
-}) || _class9);
+}) || _class8);
 
 const checkedArrayContext = 'CheckedObserver:array';
 const checkedValueContext = 'CheckedObserver:value';
 
-export let CheckedObserver = (_dec8 = subscriberCollection(), _dec8(_class10 = class CheckedObserver {
+export let CheckedObserver = (_dec8 = subscriberCollection(), _dec8(_class9 = class CheckedObserver {
   constructor(element, handler, observerLocator) {
     this.element = element;
     this.handler = handler;
@@ -3852,11 +3852,11 @@ export let CheckedObserver = (_dec8 = subscriberCollection(), _dec8(_class10 = c
       this.valueObserver.unsubscribe(checkedValueContext, this);
     }
   }
-}) || _class10);
+}) || _class9);
 
 const selectArrayContext = 'SelectValueObserver:array';
 
-export let SelectValueObserver = (_dec9 = subscriberCollection(), _dec9(_class11 = class SelectValueObserver {
+export let SelectValueObserver = (_dec9 = subscriberCollection(), _dec9(_class10 = class SelectValueObserver {
   constructor(element, handler, observerLocator) {
     this.element = element;
     this.handler = handler;
@@ -4016,7 +4016,7 @@ export let SelectValueObserver = (_dec9 = subscriberCollection(), _dec9(_class11
       this.arrayObserver = null;
     }
   }
-}) || _class11);
+}) || _class10);
 
 export let ClassObserver = class ClassObserver {
   constructor(element) {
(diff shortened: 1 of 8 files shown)


michaelw85:
@bigopon Here's a small PR to fix the shadow dom boundry event bubbling issue

src/event-manager.js
```diff
@@ -86,8 +86,8 @@ function handleDelegatedEvent(event) {
         }
       }
     }
-
-    target = target.parentNode;
+    
+    target = (target.parentNode && target.parentNode.toString() === "[object ShadowRoot]") ? target.parentNode.host : target.parentNode;
```
bigopon: Can we avoid accessing `parentNode` multiple times? It would be tough on minification, and also perf

michaelw85: @bigopon Pushed a change, caching the parentNode

src/event-manager.js
```diff
@@ -87,7 +87,10 @@ function handleDelegatedEvent(event) {
       }
     }
 
-    target = target.parentNode;
+    const parent = target.parentNode;
+    const parentIsShadowRoot = parent && parent.toString() === '[object ShadowRoot]';
+
+    target = parentIsShadowRoot ? parent.host : parent;
```
bigopon: @EisenbergEffect For your awareness, this aligns our delegation strategy with the way it works natively with trigger, but could be a potential breaking change. One way to avoid it is to add a flag on event manager, but depends on how we would look at it, since it's an app wide configuration, and one plugin requiring it would enforce this on the rest

test/event-manager.spec.js
```diff
@@ -120,6 +128,20 @@ describe('EventManager', () => {
       threeClick.calls.reset();
     });
 
+    it('bubbles properly out of shadow dom when not delegated with composed flag', () => {
```
bigopon: tests are good 👍

michaelw85: Thanks, I also corrected a potential pitfall. Moved the creation of the elements into the before each, the previous setup would append the same elements over and over again. If the tests would have checked on the amount of delegate calls I would expect it to have failed with the previous setup.

src/event-manager.js
michaelw85: I agree it could impact existing setups but only if you use shadow dom of course and have elements within shadow dom firing events. Anyone who is using such a setup probably used a workaround like I did.

I currently implemented 2 different workarounds to be able to use delegate on custom components using shadow dom. One workaround is to add an element in the light dom and trigger an event on this element when the shadow dom variant is clicked for example, this acts as a proxy. This scenario might end up producing 2 events now unless you stop the event bubbling than it would still work as it used to. 
The other workaround I used is capture the original event in the shadow dom and fire a custom event on the shadow host. This scenario would still function as expected and should not be impacted.

test/event-manager.spec.js
michaelw85: Thanks 😎

src/event-manager.js
bigopon: That sounds reasonable to me. Maybe we could just have it this way

michaelw85: FYI in my project I just confirmed that implementation 1 will now fire duplicate events and when I remove the "proxy light dom" the click event is now triggered as expected!

EisenbergEffect:
We definitely need this in some form. I had to implement some custom stuff for my own vNext app because this doesn't work quite right with link clicks and routing, for example. So, I think we need a fix in the core and also some fix in the router....and we need to port it all to vNext. Two thoughts:

* For vCurrent, I do think we should have a setting to turn this on, just to be safe. For vNext, it should just work this way by default.
* Is there any other way to detect the shadow root other than `toString`? I thought there was an API for that.

/cc @fkleuver for vNext core and @jwx for vNext router.

michaelw85:
@EisenbergEffect Where would such a flag be implemented? Would this be something global?

On the alternatives, lets brainstorm:
1) I could use `instanceof ShadowRoot`
2) I could check `element.shadowRoot` but this fail in closed mode. https://developer.mozilla.org/en-US/docs/Web/API/Element/shadowRoot
3) I could use `getRootNode()` to get the shadow root but not sure how to determine it's not returning the root dom I guess I would still end up using toString or instanceof. Another drawback I doubt this works cross browser: https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode
4) I could attempt to find the element in the light dom, if not its in shadow dom. (does not sound performant and wonder if it might end up giving false results)
5) Maybe it would be possible to match host with the dom root if it doesn't match we have successfully detected a shadow root host and should continue up from there.
6) ....

I think instanceof might be the only feasible and more elegant option .

EisenbergEffect:
It could be implemented as a boolean property on the `EventManager` class itself. I guess doing an instanceof check is the best, though we would need to guard against that not existing, in the case someone is using a browser without native support.

bigopon:
I think the combination of `.getRootNode()` and `.ownerDocument` would be sufficient. Though this increases the amount of API we use in core framework, not that it matters for vCurrent since we only have 1 runtime for web. But it could be worth it to be aware of.

cc @fkleuver for awareness

fkleuver:
`.getRootNode()` has been around since the beginning of shadow dom, so wherever shadow dom is safe to use, that piece of the api is safe to use as well.

michaelw85:
I've changed the string check to an `instanceof` check. This seems like the easiest most readable change to me.

I need some assistance for the boolean/setting.
How would the `handleDelegatedEvent` function get access to this bool property? 
Should it be on the proto chain of the `EventManager` class? Or would this bool have to be passed through multiple classes/functions to end up in the `handleDelegatedEvent` function as a param? Is there a singleton I could get and should this prop just be a public property I can access?

How would a dev set this bool? 
If the `EventManager` is a singleton you could gain access using dep injection and than set the value of the bool. If its a public prop or on the proto chain this would be very fragile. Any piece of code (or plugin for that matter) could affect this value causing unexpected behavior.

I'm a bit confused how to properly pass the setting from an app all the way down to the code handling the event bubbling.

bigopon:
I think instanceof check combined with a existance check is pretty ok. About the bool check, let me have a look tonight

michaelw85:
@bigopon Did you have time to have a look at how I could introduce a flag? It would be awesome if I could wrap up this PR.

src/event-manager.js
```diff
@@ -87,7 +87,10 @@ function handleDelegatedEvent(event) {
       }
     }
 
-    target = target.parentNode;
+    const parent = target.parentNode;
+    const parentIsShadowRoot = parent && parent instanceof ShadowRoot;
```
bigopon: @michaelw85 Sorry for this, but I think your previous resolution is better. on browsers without shadow DOM, this will throw as ShadowRoot is undefined. Unless we do a small check on global.

bigopon:
We have
```js

class DelegateHandlerEntry {
  constructor(eventName) {
    this.eventName = eventName;
    this.count = 0;
  }

  increment() {
    this.count++;

    if (this.count === 1) {
      DOM.addEventListener(this.eventName, handleDelegatedEvent, false);
    }
  }

  decrement() {
    if (this.count === 0) {
      emLogger.warn('The same EventListener was disposed multiple times.');
    } else if (--this.count === 0) {
      DOM.removeEventListener(this.eventName, handleDelegatedEvent, false);
    }
  }
}
```

We can change the constructor to accept 1 more parameter, that it will determine if it should check for shadow root, and then we can change handler from a loose function, to event listener object:
```js

class DelegateHandlerEntry {
  constructor(eventName, escapeShadowRoot) {
    this.eventName = eventName;
    this.count = 0;
    this.escapeShadowRoot = escapeShadowRoot;
  }

  handleEvent(event) {
    // this will be this instance
    // so we can get the escapeShadowRoot value
  }

  increment() {
    this.count++;

    if (this.count === 1) {
      // use this as handler, instead
      DOM.addEventListener(this.eventName, this, false);
    }
  }

  decrement() {
    if (this.count === 0) {
      emLogger.warn('The same EventListener was disposed multiple times.');
    } else if (--this.count === 0) {
      // use this to remove handler instead
      DOM.removeEventListener(this.eventName, this, false);
    }
  }
}
```

Though will need some more tests for these changes. Are you comfortable with writing tests for those changes?

src/event-manager.js
fkleuver: I like the `instanceof` though. Could do this for safety:

```ts
const parentIsShadowRoot = typeof ShadowRoot !== 'undefined' && parent instanceof ShadowRoot;
```

No need for the truthy check, that's implied by `instanceof`

michaelw85:
@bigopon Thanks for the input but I'm still a bit confused how an Aurelia app would pass this bool to the constructor. How does the consuming app set this to `true`?

bigopon:
@michaelw85 During start up, one would do this:
```js
aurelia.container.get(EventManager).escapeShadowRoot = true;
```

then we change the code here: https://github.com/aurelia/binding/blob/b42630b9ef94f84f39e450d959ddaa721d82e5d5/src/event-manager.js#L280
to
```js
    this.defaultEventStrategy = new DefaultEventStrategy(this);
```

So later, we could change here https://github.com/aurelia/binding/blob/b42630b9ef94f84f39e450d959ddaa721d82e5d5/src/event-manager.js#L180 and here https://github.com/aurelia/binding/blob/b42630b9ef94f84f39e450d959ddaa721d82e5d5/src/event-manager.js#L200 to something like this:

```diff
-      handlerEntry = capturedHandlers[targetEvent] || (capturedHandlers[targetEvent] = new CapturedHandlerEntry(targetEvent));
+      handlerEntry = capturedHandlers[targetEvent] || (capturedHandlers[targetEvent] = new CapturedHandlerEntry(targetEvent, this.eventManager.escapeShadowRoot));
```

michaelw85:
ok thanks I could give this a try.
Not a fan of getting the instance and changing the prop tbh, this could be done from anywhere by any code causing unexpected side effects/breaking the code. What if for instance a plugin would do this?

bigopon:
@michaelw85 after `aurelia.start()` call, all plugins have been run, you can set it there. About your Q, there's always possibility of something messing with this, which I don't see anything we can do about, but I'm not sure anyone would do that unless for fun. Beside that, as you said, it matches native behavior, pretty ok to me

michaelw85:
Only option I could think of as an alternative is to only read the option once at startup from the aurelia config. This would prevent any changes during run time.

bigopon:
@michaelw85 we can do it that way by having the setter ignore 2nd and after assignment.

Cc @EisenbergEffect @fkleuver for making a call with us

EisenbergEffect:
I'm pretty ok with keeping it simple and letting people do this"

```JavaScript
aurelia.container.get(EventManager).escapeShadowRoot = true;
```

While there is technically a possibility for things to stomp on each other, it seems unlikely.

src/event-manager.js
michaelw85: With refactoring the original behavior will not hit the instanceof check anymore.

michaelw85:
@bigopon Sorry it took a while I had to finish some other work first but here it is.
Implemented the flag, struggled a bit with unit tests but got it up and running after disposing the registered events after each test 🌈

michaelw85:
@EisenbergEffect This PR is still waiting for review, could someone be assigned?

bigopon:
Oops, seems like its my turn to slack. Ill review it tonight. Sorry for that. First glance is it needs some modification

src/event-manager.js
```diff
@@ -67,49 +67,53 @@ class CapturedHandlerEntry {
...
-      if (callback) {
-        if (event.stopPropagation !== stopPropagation) {
-          event.standardStopPropagation = event.stopPropagation;
-          event.stopPropagation = stopPropagation;
-        }
-        if ('handleEvent' in callback) {
-          callback.handleEvent(event);
-        } else {
-          callback(event);
+  handleDelegatedEvent = event => {
```
bigopon: This means every entry, we will create a new fn for handling the event. I don't think this is the direction we would want to go.
So it looks like we need a way to associate the handler with an Entry instance, `handleEvent` seems like a perfect interface for this.

So we can have this:
```diff
class DelegateHandlerEntry {

+  handleEvent(event) {
+    handleDelegatedEvent(event, this.escapeShadowDom); // or we can just inline the existing loose function handler here
+  }
-  handleDelegatedEvent = event => {
-    ...  
}
```
And please don't forget capture too 😄

michaelw85: @bigopon Changed the arrow func into a class method, as a result the handleDelegatedEvent becomes part of the prototype chain. This means only the params in the this scope will be stored in the new object, so overhead on memory is minimized now.

dist/es2015/aurelia-binding.js
```diff
@@ -3078,51 +3078,53 @@ let CapturedHandlerEntry = class CapturedHandlerEntry {
...
   }
 };
+let DelegateHandlerEntry = class DelegateHandlerEntry {
+  constructor(eventName, eventManager) {
+    this.eventName = eventName;
+    this.count = 0;
+    this.eventManager = eventManager;
+  }
 
+  handleDelegatedEvent(event) {
```
bigopon:
```suggestion
  handleEvent(event) {
```
And then in `increment()`:
```ts
DOM.addEventListener(this.eventName, this, false);
```
and in `decrement()`:
```ts
DOM.removeEventListener(this.eventName, this, false);
```

src/event-manager.js
```diff
@@ -67,49 +67,53 @@ class CapturedHandlerEntry {
...
-      if (callback) {
-        if (event.stopPropagation !== stopPropagation) {
-          event.standardStopPropagation = event.stopPropagation;
-          event.stopPropagation = stopPropagation;
-        }
-        if ('handleEvent' in callback) {
-          callback.handleEvent(event);
-        } else {
-          callback(event);
+  handleDelegatedEvent(event) {
```
bigopon:
```suggestion
  handleEvent(event) {
```
And then in `increment()`:
```ts
DOM.addEventListener(this.eventName, this, false);
```
and in `decrement()`:
```ts
DOM.removeEventListener(this.eventName, this, false);
```

src/event-manager.js
```diff
@@ -67,49 +67,53 @@ class CapturedHandlerEntry {
...
+          }
         }
       }
-    }
 
-    target = target.parentNode;
-  }
-}
+      const parent = target.parentNode;
+      const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent && parent instanceof ShadowRoot;
```
bigopon:
```suggestion
      const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent && parent instanceof ShadowRoot;
      const shouldEscapeShadowRoot = this.eventManager.escapeShadowRoot && parent instanceof ShadowRoot;
```
Would this be a `ReferenceError` in old browser? Sorry I don't remember what we decided to go with this. Or the argument is that if user turns on `escapeShadowRoot`, they should be already aware of potential error here?

michaelw85: Using shadow dom is an explicit choice of the user which would make him/her aware of limitations in old browsers so you would only turn this on when supported otherwise escape would be false by default which would make the code never reach the instanceof check.

src/event-manager.js
michaelw85: @bigopon Updated the PR. Thanks for this reply! This resolved an issue I could not wrap my mind around, I thought something was wrong with the unit tests cleaning the listeners but apparently I just made a mistake in the code itself... nice catch

bigopon:
@michaelw85 nice 👍 . Can you help give `CapturedHandlerEntry` the same treatment like `DelegatedHandlerEntry`? Also please remove the dist files

michaelw85:
> @michaelw85 nice 👍 . Can you help give `CapturedHandlerEntry` the same treatment like `DelegatedHandlerEntry`? Also please remove the dist files

Changing `CapturedHandlerEntry` seems unrelated to this PR. I would not mind changing it but I would rather create a new PR for that with a proper description why it is changed.

The dist files are not ignored at the moment should they be? Should I revert the changed of the dist in this PR or do you really mean delete? But I guess same question applies here is this related to this PR else I would prefer keeping it out of this one.

bigopon:
Well, i just realized the title of this PR, so I think your argument is fair, and in agile spirit. We probably can go ahead with this. Just need to make @EisenbergEffect @fkleuver aware of this.
For dist files, we have been going with that for a long time, I'm not sure why.

michaelw85:
Ok great lets finish this PR. If you create tasks/placeholders for the other changes just mention me and I will create PR's.

EisenbergEffect:
The `dist` files are needed for Bower, since it uses the repo itself as the installation source. We'll be dropping this for Aurelia 2, but it's not something we can drop for the current version, due to people being dependent on it. It could be solved with an imporved CI process and different branch strategy, but we've had it this way so long, and it's only a minor inconvenience, so we haven't allotted the time to make the change.

@michaelw85 If you can remove the dist folder from this PR, that would be great.
@bigopon Are we tracking this for Aurelia 2 as well? We will need to make the same improvements or we're going to hit these problems again right out of the gate.

michaelw85:
@EisenbergEffect @bigopon I've reverted the dist changes as requested.

bigopon:
@EisenbergEffect https://github.com/aurelia/aurelia/issues/325#issuecomment-527858820 done

EisenbergEffect:
Thanks @michaelw85 and @bigopon 

Doing a quick review now. This is technically a breaking change, but I somehow doubt that it's going to affect anyone.

Does anyone here have concerns? Do we need to put this behind a flag?

EisenbergEffect:
Doh. I see we did put a flag there. Just ignore me. Doing final review now...turning on my thinking cap.

PR #762, opened 2019-09-05
fix(event-manager): update d.ts

michaelw85:
Add escapeShadowRoot to d.ts

Related: #758 #755


michaelw85:
@EisenbergEffect @bigopon I just noticed that I forgot to add the newly introduced `escapeShadowRoot` flag to the d.ts file. This PR add's the flag

PR #766, opened 2019-11-12
fix(event-manager): Firefox shadow dom event handling

### LINKED ISSUE #765: Delegate does not support shadow dom for Firefox

Related to: #755 

* **Library Version:**
2.5.0

* **Operating System:**
Windows [10]

* **Node Version:**
10.15.1

* **NPM Version:**
6.10.2

* **Browser:**
Firefox 70.0.1 (64-bits)

* **Language:**
all

**Current behavior:**
Clicking nested elements in shadow dom does not trigger click delegate.

In my case this fails for a search input with a search button.
image

The search button has a click.trigger which should dispatch a custom "on-search" event. This fails due to:
```ts
// event-manager.js:7
function findOriginalEventTarget(event) {
  return (event.path && event.path[0]) || (event.deepPath && event.deepPath[0]) || event.target;
}
```

**Result in firefox:**
`spec-input-search` (custom search component containing the search button)

Chrome (which works properly):
`button`

**Reason for failure:**
Firefox does not have event.path or deepPath so falls back to target which does not take shadow dom into account. 

**Proposed solution:**
Use composedPath before using fallback to event.target

When using shadow dom, nested should handle click delegate properly

michaelw85:
Return correct original event target for Firefox

Closes #765

### DIFF
--- a/src/event-manager.js
+++ b/src/event-manager.js
@@ -5,7 +5,7 @@ import * as LogManager from 'aurelia-logging';
 let emLogger = LogManager.getLogger('event-manager');
 //Note: path and deepPath are designed to handle v0 and v1 shadow dom specs respectively
 function findOriginalEventTarget(event) {
-  return (event.path && event.path[0]) || (event.deepPath && event.deepPath[0]) || event.target;
+  return (event.path && event.path[0]) || (event.deepPath && event.deepPath[0]) || (event.composedPath && event.composedPath()[0]) || event.target;
 }
 
 function stopPropagation() {
--- a/src/event-manager.js
+++ b/src/event-manager.js
@@ -5,7 +5,7 @@ import * as LogManager from 'aurelia-logging';
 let emLogger = LogManager.getLogger('event-manager');
 //Note: path and deepPath are designed to handle v0 and v1 shadow dom specs respectively
 function findOriginalEventTarget(event) {
-  return (event.path && event.path[0]) || (event.deepPath && event.deepPath[0]) || (event.composedPath && event.composedPath()[0]) || event.target;
+  return (event.composedPath && event.composedPath()[0]) || (event.deepPath && event.deepPath[0]) || (event.path && event.path[0]) || event.target;
 }
 
 function stopPropagation() {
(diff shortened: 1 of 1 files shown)


michaelw85:
@bigopon @fkleuver Followup for #755

fkleuver:
@bigopon @michaelw85 Do we have any particular tests for this? In v2 I did this a little bit differently, starting with the more specific APIs first and falling back to the more generic ones. I seem to recall discussion that there was something wrong about v2's event handling but not sure if it applied to this.
Should this code be changed to be in sync with v2? https://github.com/aurelia/aurelia/blob/master/packages/runtime-html/src/observation/event-manager.ts#L16
Or should it be the other way around you think? @bigopon 
Either way I'd prefer if we kept v1 and v2 the same in this particular regard.

michaelw85:
@fkleuver I tested the fix locally in my project (created a branch with updated dist). The method is private and I did not think any tests would run in Firefox so I did not write a unit test. I could of course quite easily provide a simple component that could be used to verify the fix.

On the approach, I would prefer the v2 setup. This seems correct to me, I thought the composedPath is the "new" standardized api for getting the target info. I was thinking about this approach with my change but decided to minimize impact. if @bigopon agrees I can update the PR to match v2

bigopon:
If `path` and `composedPath` are not the same in chrome, then we cannot move `composedPath` in front of `path`, unless we have a way to configure it in the event manager

michaelw85:
`path` should be the same, apparently this was the initial proposal by Google and this became `composedPath()`. `composedPath` is the standard now. 

A quick test in my app shows they are identical in this scenario:
image

bigopon:
@michaelw85 thanks for checking, If so let's go with @fkleuver 's suggestion and make it behave like v2

michaelw85:
ok will update.

@fkleuver I just spotted a mistake in the v2 code. `event.deepPath()[0]` deepPath is an array not a method.
Documentation: https://developer.mozilla.org/en-US/docs/Web/API/Event
```
Event.deepPath 
An Array of DOM Nodes through which the event has bubbled.
```

michaelw85:
@bigopon @fkleuver Can this PR be reviewed or are you waiting for me to do any additional changes?

fkleuver:
I think we're good. Thank you!

@EisenbergEffect Unless you see any issue, as far as I'm concerned we can merge/release this. It's now in sync with v2.

EisenbergEffect:
Thanks everyone! We'll get this out next week.

PR #772, opened 2020-04-24
Unset oldValue after being done with it

### LINKED ISSUE #771: SetterObserver keeps old value retained

* **Library Version:**
2.2.0

* **Browser:**
all

**Current behavior:**
SetterObserver keeps a reference to the old value of a property, and doesn't release it until the property is changed a subsequent time. This can cause objects to retained much longer than needed.

SetterObserver releases the old value as soon as it's done with it.

ccantill:
fixes #771

### DIFF
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -75,6 +75,8 @@ export class SetterObserver {
     this.queued = false;
 
     this.callSubscribers(newValue, oldValue);
+
+    this.oldValue = null;
   }
 
   subscribe(context, callable) {
--- a/src/property-observation.js
+++ b/src/property-observation.js
@@ -71,12 +71,11 @@ export class SetterObserver {
   call() {
     let oldValue = this.oldValue;
     let newValue = this.currentValue;
+    this.oldValue = null;
 
     this.queued = false;
 
     this.callSubscribers(newValue, oldValue);
-
-    this.oldValue = null;
   }
 
   subscribe(context, callable) {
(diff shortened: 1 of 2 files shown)


bigopon:
@ccantill thanks. I see what you meant. I'll have a look to see whether we need to be carefully setting initial value of old value to something else or simply null would suffice

bigopon:
I think this is fine. As the observer always deals with outside using `currentValue`.
We may need this for v2 as well cc @fkleuver @EisenbergEffect

fkleuver:
Actually I already solved this in v2 (unintentionally) when I was doing perf optimizations: <https://github.com/aurelia/aurelia/blob/master/packages/runtime/src/observation/setter-observer.ts#L61>

I think we can do it the same way in v1; just change this:
```ts
  call() {
    let oldValue = this.oldValue;
    let newValue = this.currentValue;
    this.oldValue = null;

    this.queued = false;

    this.callSubscribers(newValue, oldValue);
  }
```

To this:
```ts
  call() {
    let oldValue = this.oldValue;
    let newValue = this.oldValue = this.currentValue;

    this.queued = false;

    this.callSubscribers(newValue, oldValue);
  }
```

It won't matter much, but it can be slightly better for perf in otherwise monomorphic hot paths.

bigopon:
That looks good to me, I was thinking of that too but wouldn't want to bother too much with non-outcome requests 😛 
@ccantill can you please help with that?

ccantill:
I've changed it in my PR.

ccantill:
@bigopon any idea on when this can be merged/released?

bigopon:
@ccantill oops, sorry missed this. pinging @fkleuver @EisenbergEffect for a merge & release

fkleuver:
@ccantill It's released now. Thank you!

Issue #795, opened 2025-07-25
`AccessMember.evaluate` should always return `undefined` for non-existing members


nenadvicentic:
**The Issue**

Current version of `aurelia-bindinig@2.5.5` resolves value of non-existing member of an object differently when the object is `null` vs `undefined` or non existant.

For example, binding expression `foo.bar` would resolve to `null` for:

```js
{ foo: null }
```
but would resolve to `undefined` for:
```js
{ foo: undefined }
```
and
```js
{}
```

while if `foo` is an empty object, binding expression `foo.bar` would correctly return `undefined`:
```js
{ foo: {} }
```

This unpredictability produces issue in some binding scenarios. For example:
```html
<input type="radio" model.bind="null" checked.bind="foo.bar" />
<input type="radio" model.bind="false" checked.bind="foo.bar" />
<input type="radio" model.bind="true" checked.bind="foo.bar" />
```

The culprit is this method:
https://github.com/aurelia/binding/blob/ed35588117f52508e628570921bc2361c3861cc7/src/ast.js#L251-L254

Instead of returning `instance`, it should return `undefined`:
```js
evaluate(scope, lookupFunctions) {
    let instance = this.object.evaluate(scope, lookupFunctions);
    return instance === null || instance === undefined ? undefined: instance[this.name];
  }
```
**Expected behavior**

`foo.bar` binding expression should return `undefined` for all 4 cases:

```js
{ foo: null }
{ foo: undefined }
{ foo: {} }
{}
```
It would also be consistant with result of JavaScript optinal chaning syntax `this.foo?.bar`.

PR #796, opened 2025-07-25
fix: AccessMember.evaluate should always return undefined for non-existing members #795

nenadvicentic:
Fix for issue the #795

### DIFF
--- a/dist/amd/aurelia-binding.js
+++ b/dist/amd/aurelia-binding.js
@@ -1504,7 +1504,7 @@ define(['exports', 'aurelia-logging', 'aurelia-pal', 'aurelia-task-queue', 'aure
 
     AccessMember.prototype.evaluate = function evaluate(scope, lookupFunctions) {
       var instance = this.object.evaluate(scope, lookupFunctions);
-      return instance === null || instance === undefined ? instance : instance[this.name];
+      return instance === null || instance === undefined ? undefined : instance[this.name];
     };
 
     AccessMember.prototype.assign = function assign(scope, value) {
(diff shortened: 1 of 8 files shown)


bigopon:
Thanks @nenadvicentic , this is probably the most uncomfortable issue I had with v1 behavior. While I agree we wont want to randomly return null/undefined, i dont think we can just change the behavior without an optin mechanism. In v2 we have the opportunity to rectify all this without any issues but v1 will need some effort.

nenadvicentic:
@bigopon I get what your concern is, but I think this edge case is very rare, therefore we are fixing a glitch, not chaning the behaviour.

On top of that, it's very unstable edge case. It occurs only when parent object is `null` and we are making strict `undefined` or `null` check on a **direct** property of that object. For parent to be `null`, it has to be asigned by application (code or API call).

But more importantly, the moment end-user sets any other child property of `null` parent via data binding, Aurelia asignes new object to that parent. In that moment my missing property turns from `null` to `undefined`. It's unpredictable behaviour that developer has to shield against . I cannot see any use-case where someone actually relly on such glitch.

In addition, I run all 1795 already existing unit tests after applying this change. Everything was green, despite my change.

Btw, there is 2nd pull request where test runner is working with Node 22.

nenadvicentic:
@bigopon Let me express myself in a programmers language...  :)

To work around this glitch, I have to relax my checks of `parent.property`. From:

```js
parent?.property === undefined
```
to:
```js
parent?.property === undefined || parent?.property === null
```
I coudn't have used:
```js
parent?.property === null
```
because property turns into `undefined` when parent is not `null` object any more.

So, framework constraining `parent?.property` to always be `undefined` instead of unpredicable `undefined | null`, won't break any developer code.

